PR #6, opened 2018-11-29
Integrating gitdata and misc fixes

adammhaile:
@tbielawa @sosiouxme @smunilla @joeldavis84 
Another kinda hefty change.
Setting up for being able to have ocp-build-data be branched (which I've already tested works with a 2 line change, currently commented out with note).
This mainly streamlines the data loading a lot making it easier for Elliott to use the same base data load code but eventually do away with runtime entirely. (Really elliott only needs the name from the yaml)
Also made a bunch of small fixes.
And I removed `config:new` because it really was problematic as a workflow. I added `config:get` instead so that there's a one-off for "give me the latest config local", which you can then edit and use `config:push` to update.

### DIFF
--- a/doozer
+++ b/doozer
@@ -10,7 +10,7 @@ from doozerlib.config import MetaDataConfig as mdc
 from doozerlib.config import valid_updates
 from doozerlib import cli_opts
 from doozerlib.exceptions import DoozerFatalError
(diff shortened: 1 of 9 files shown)


tbielawa:
/lgtm
ship it

PR #8, opened 2018-11-30
Give doozer a top level --version option

tbielawa:


adammhaile:
Nice, but I would like to consolidate where version is stored. It should read that value from somewhere that setup.py can get to as well so we only have to change it in one place. I have some example code of how I've done it before... will come back to this.

adammhaile:
Also, yes we should make it 0.2.0 or whatever. I meant to do that.

tbielawa:
Glad you agree, wrt: where it's stored.

https://click.palletsprojects.com/en/6.x/api/#click.version_option

```
Parameters
version – the version number to show. If not provided Click attempts an auto discovery via setuptools.
```

They just don't tell us HOW it tries to figure that out. When I tried using it implicitly it failed :'(

adammhaile:
Ok, I'll take a look.

tbielawa:
Superseded by #12.

PR #9, opened 2018-11-30
Fix stream load

adammhaile:


tbielawa:
/lgtm
not that i know what it is. but it looks good to me.

PR #11, opened 2018-12-03
minor tweaks

sosiouxme:
mostly, let you know what command is not available when doozer tries to run something that's not there

@adammhaile any concerns?


adammhaile:
@sosiouxme I keep forgetting there's no auto-merge here. Merge when ready.

PR #12, opened 2018-12-03
Adding better version handling

adammhaile:
@tbielawa I was never able to get #8 working even with my vast knowlege of setuptools :( I've looked at the click code and it makes no sense. So I've implemented it the way I always do it which is with a `VERSION` file in the module that gets read by setup.py and doozer itself. The reason for the version fetch function sort of repeated in both places is that I can't just have setup.py load doozer as it would lack the dependencies at that point. A necessary evil.


doozer
```diff
@@ -31,10 +32,29 @@ pass_runtime = click.make_pass_decorator(Runtime)
 context_settings = dict(help_option_names=['-h', '--help'])
 
 
+VERSION_QUOTE = """
+The Doozers don't mind their buildings being eaten;
+if the Fraggles didn't eat the constructions,
+the Doozers would run out of building space,
+and if they ran out of building space,
+they would have to move away from Fraggle Rock
+or else they would die.
```
tbielawa: You're the best

doozer
```diff
@@ -31,10 +32,29 @@ pass_runtime = click.make_pass_decorator(Runtime)
...
+"""
+
+
+def print_version(ctx, param, value):
+    if not value or ctx.resilient_parsing:
+        return
+    click.echo('Doozer v{}'.format(version()))
+    click.echo(VERSION_QUOTE)
```
tbielawa: For the lols. This is great.

doozer
adammhaile: It was either this or an ASCII doozer face

PR #14, opened 2018-12-06
Fixing symlink and ignore missing base issues

adammhaile:
@joeldavis84 This fixes that telemeter symlink issue. Turns out we've never had a container repo with a symlink in the root. Oops.


tbielawa:
lgtm

PR #15, opened 2018-12-06
Experimental mirror input and ImageStream generator

tbielawa:
Generates input file for 'oc image mirror' command. Used to
synchronize OCP content from internal registries to quay.

Generates a kubernetes ImageStream object to send to a release
controller which is responsible for creating update and install
payloads.

This is still experimental and VERY specific to ART
requirements. Requires future refactoring to be generally usable.

### DIFF
--- a/doozer
+++ b/doozer
@@ -899,8 +899,7 @@ def images_print(runtime, short, show_non_release, show_base, output, pattern):
         try:
             dfp.content = image.fetch_cgit_file("Dockerfile")
         except Exception:
-            click.echo("Error reading Dockerfile from distgit: {}".format(image.distgit_key))
-            raise
+            raise DoozerFatalError("Error reading Dockerfile from distgit: {}".format(image.distgit_key))
 
         version = dfp.labels["version"]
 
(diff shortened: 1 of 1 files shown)


doozer
```diff
@@ -1204,6 +1204,160 @@ def config_print(runtime, key, name_only):
     config.config_print(key, name_only)
 
 
+@cli.command("beta:release-gen", short_help="Generate input files for release mirroring")
```
adammhaile: I <3 the `beta` option here. Great way to add experimental stuff :)

doozer
```diff
@@ -1204,6 +1204,160 @@ def config_print(runtime, key, name_only):
...
+@click.option("--src-dest", "--sd", default="img-mirror.txt",
+              help="The SRC=DEST file to write to\ndefault=img-mirror.txt")
+@click.option("--image-stream", "--is", default="release-is.yaml",
+              help="The SRC=DEST file to write to\ndefault=release-is.yaml")
```
adammhaile: `SRC=DEST` looks like it's a copy-paste error?

doozer
```diff
@@ -1204,6 +1204,160 @@ def config_print(runtime, key, name_only):
...
+
+    # All the items we will put in the image mirror input
+    src_dest_items = []
+
+    for image in images:
+        dfp = DockerfileParser(path=runtime.working_dir)
+        try:
+            dfp.content = image.fetch_cgit_file("Dockerfile")
+        except Exception:
+            click.echo("Error reading Dockerfile from distgit: {}".format(image.distgit_key))
```
adammhaile: simply `raise DoozerFatalError(msg)` here instead. This will be caught at top level by doozer and instead of dropping a nasty stacktrace, will print the error in red and then `sys.exit(1)` cleanly.

doozer
```diff
@@ -1204,6 +1204,160 @@ def config_print(runtime, key, name_only):
...
+        s = s.replace("{pushes}", '{}\n'.format(pushes_formatted))
+
+        if "{" in s:
+            raise IOError("Unrecognized fields remaining in pattern: %s" % s)
+
+        # Per clayton:
+        """Tim Bielawa: note to self: is only for `ose-` prefixed images
+        Clayton Coleman: Yes, Get with the naming system or get out of town
+        """
+        if 'ose' in image.image_name_short and '666' not in release:
```
adammhaile: lol at 666. I'm sorry timmy. 
This should be something that eventually gets added to a config or passed in at the CLI. I will allow it now because I know this is experimental, but eventually of course we'll remove anything specific to us.

doozer
```diff
@@ -1204,6 +1204,160 @@ def config_print(runtime, key, name_only):
...
+                    'name': dest
+                }
+            })
+        else:
+            red_prefix("NOT adding to IS: ")
+            print(s)
+
+        # This indicates this is an ART test build, it has not been
+        # mirrored yet. Don't try to sync it across registries.
+        if '666' not in release:
```
adammhaile: Same comment as above.

doozer
tbielawa: oh crap, lol

tbielawa: I wonder if I meant to leave that there. Oh well. I'll follow your suggestion and see how it goes.

PR #16, opened 2018-12-13
distgit builds: analyze logs ART-392 ART-366

sosiouxme:
check build logs for package install failures and multiarch component
mismatches.

ART-392 Fail faster on known build problems
When build logs indicate failures that we recognize and don't expect to
be solved by retries, the build is not retried.

ART-366 Detect missing RPMs in build logs
Even for successful builds, the logs are checked for problems. At this
time, that's just packages that didn't actually install, but could
expand.

Here these checks look for the same problems. If it becomes
necessary then there could easily be two separate problem lists.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -198,8 +198,8 @@ class DistGitRepo(object):
 
 
 class ImageDistGitRepo(DistGitRepo):
-    def __init__(self, metadata):
-        super(ImageDistGitRepo, self).__init__(metadata)
+    def __init__(self, metadata, autoclone=True):
+        super(ImageDistGitRepo, self).__init__(metadata, autoclone)
         self.build_lock = Lock()
         self.build_lock.acquire()
         self.logger = metadata.logger
@@ -628,6 +628,13 @@ class ImageDistGitRepo(DistGitRepo):
 
                 def wait(n):
                     self.logger.info("Async error in image build thread [attempt #{}]".format(n + 1))
+                    # No need to retry if the failure will just recur
+                    error = self._detect_permanent_build_failures(self.runtime.group_config.image_build_log_scanner)
+                    if error is not None:
+                        raise exectools.RetryException(
+                            "Saw permanent error in build logs:\n{}\nWill not retry after {} failed attempt(s)"
+                            .format(error, n + 1)
+                        )
                     # Brew does not handle an immediate retry correctly, wait
                     # before trying another build, terminating if interrupted.
                     if terminate_event.wait(timeout=5 * 60):
@@ -752,11 +759,13 @@ class ImageDistGitRepo(DistGitRepo):
             error = None
 
         # Gather brew-logs
-        logs_dir = "%s/%s" % (self.runtime.brew_logs_dir, self.metadata.name)
-        logs_rc, _, logs_err = exectools.cmd_gather(["brew", "download-logs", "-d", logs_dir, task_id])
+        logs_rc, _, logs_err = exectools.cmd_gather(["brew", "download-logs", "-d", self._logs_dir(), task_id])
 
         if logs_rc != 0:
             self.logger.info("Error downloading build logs from brew for task %s: %s" % (task_id, logs_err))
+        elif error is None:
+            # even if the build completed without error, check the logs for problems
+            error = self._detect_permanent_build_failures(self.runtime.group_config.image_build_log_scanner)
 
         if error is not None:
             # An error occurred. We don't have a viable build.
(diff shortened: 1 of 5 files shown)


sosiouxme:
We may want to wait for ART-391 to be complete before deploying this. Or we can just let jenkins base builds fail until fixed.

There may also be other install failures that come to light because of this, of course.

doozerlib/distgit.py
```diff
@@ -766,6 +776,43 @@ def _build_container(
...
+            if not task_dirs:
+                self.logger.info("No task logs found under {}; cannot analyze logs".format(logs_dir))
+                return None
+            latest_dir = max(task_dirs, key=os.path.getmtime)
+
+            # check the log files for recognizable problems
+            found_problems = []
+            problem_regexen = [
+                # note: we read log lines, so these can only match in a single line
+                r"No package .* available",
```
adammhaile: Ideally these should be in a separate file. This list will likely get very big real quick. So either a yaml file or just a python file with nothing but a list in it (I'd lean towards the latter.)

doozerlib/distgit.py
```diff
@@ -766,6 +776,43 @@ def _build_container(
...
+                r"Failed component comparison for components:.*",
+            ]
+            for log_file in os.listdir(latest_dir):
+                if log_file not in ["x86_64.log", "ppc64le.log", "s390x.log", "aarch64.log", "task_failed.log", "orchestrator.log"]:
```
adammhaile: This list could also go in above mentioned file.

doozerlib/distgit.py
adammhaile: I agree we can add that feature later if we need it, but I still want a separate file. Otherwise I fear it might turn into the same thing as "let's just put the sources in group.yaml, there aren't that many!"

tbielawa: I'm thinking about other teams. Anything hard coded might not be useful to another team. Seems like these should always come from a separate input file.

adammhaile:
Feature creep plan... Put it in ocp-build-data

On Dec 13, 2018 2:42 PM, "Tim Bielawa" <notifications@github.com> wrote:

*@tbielawa* commented on this pull request.
------------------------------

In doozerlib/distgit.py
<https://github.com/openshift/doozer/pull/16#discussion_r241536532>:

> +        """check logs to determine if this container build failed for a known non-flake reason"""
+        logs_dir = self._logs_dir()
+        try:
+            # find the most recent subdir, which should contain logs
for the latest build task
+            task_dirs = [os.path.join(logs_dir, d) for d in
os.listdir(logs_dir)]
+            task_dirs = [d for d in task_dirs if os.path.isdir(d)]
+            if not task_dirs:
+                self.logger.info("No task logs found under {}; cannot
analyze logs".format(logs_dir))
+                return None
+            latest_dir = max(task_dirs, key=os.path.getmtime)
+
+            # check the log files for recognizable problems
+            found_problems = []
+            problem_regexen = [
+                # note: we read log lines, so these can only match in
a single line
+                r"No package .* available",

I'm thinking about other teams. Anything hard coded might not be useful to
another team. Seems like these should always come from a separate input
file.

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/doozer/pull/16#discussion_r241536532>, or mute
the thread

doozerlib/distgit.py
sosiouxme: Hm. Maybe it wouldn't be hard to have this all in group.yml.

sosiouxme:
On Thu, Dec 13, 2018 at 3:23 PM Adam Haile <notifications@github.com> wrote:

> No! New file. Group.yml is too cluttered
>
>
New file under ocp-build-data?

tbielawa:
Yes

adammhaile:
Yep

sosiouxme:
SUP DAWG I HERD U DON'T LIEK LITERAL LISTS IN YO CODE SO I ACCIDENTALLY IN YO GROUP.YML

Seriously, I think that is the best place for defining something used for all doozer builds. I do not want to remember that something is off in some side file, load it as a separate object, validate it separately, etc. -- when every build in the group is affected by it, like the rest of group.yml

Corresponding build data for what was previously in code:
https://github.com/sosiouxme/ocp-build-data/commit/b81ff59e6aafd58f46d257a52f610fbdeeed03de

Thus log scanning is initially inactive and can be rolled out and rolled back without further doozer changes.

adammhaile:
@sosiouxme it's a number of things. One is certainly clutter. Another is separation of tools. @smunilla is already doing this with multiple yaml files for elliott's various data needs. Another is that I personally see a potential for that list to get mighty big eventually and I don't want to move it later.
But how about this: I've expressed my concerns and will call that enough for now. I mostly wanted to make sure we thought through all the options. I agree that it's a simple change (at the moment) and don't want to hold up the rapid progression of Doozer. So let's ship it.

sosiouxme:
:ship: :shipit:

PR #17, opened 2018-12-13
We need to make errors like this wayyyyyyyyyyyyyy easier to debug.

tbielawa:


doozerlib/distgit.py
```diff
@@ -1135,8 +1135,15 @@ def _merge_source(self):
             if os.path.isfile("Dockerfile"):
                 os.remove("Dockerfile")
 
-            # Rename our distgit source Dockerfile appropriately
-            os.rename(dockerfile_name, "Dockerfile")
+            try:
+                # Rename our distgit source Dockerfile appropriately
+                os.rename(dockerfile_name, "Dockerfile")
+            except OSError as err:
+                print("TODO: Print out what dang image this is and a link to the source on github, if that's not too hard.")
```
tbielawa: and by `source` i mean `directory`

doozerlib/distgit.py
```diff
@@ -1135,8 +1135,15 @@ def _merge_source(self):
...
+                print("Expected dockerfile '{}' does not exist, but these do!")
```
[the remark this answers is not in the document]
tbielawa: This would help at the very least
```suggestion
                print("Expected dockerfile '{}' does not exist, but these do!".format(os.path.realpath(os.curdir)))
```

doozerlib/distgit.py
```diff
@@ -1135,8 +1135,15 @@ def _merge_source(self):
...
+                dir_contents = os.listdir('.')
+                print("\n".join(dir_contents))
```
sosiouxme: shouldn't this look for names similar to `Dockerfile`?

doozerlib/distgit.py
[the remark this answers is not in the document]
adammhaile: `self.distgit_dir` is what you should use. Nothing should ever be dependent on `pwd` here.

doozerlib/distgit.py
adammhaile: No `print()` please! Use logger.

tbielawa:
Adam implemented this in a different PR and hit a few other spots as well.

PR #18, opened 2018-12-19
[WIP] Adding local builds

adammhaile:


sosiouxme:
OK, cool. This is gonna be a great option. Please read the following as constructive :) This will likely be a user's first view of this tool and it's still pretty heavily geared toward more sophisticated usage / actual brew usage, could use some ~~dumbing down~~ UX improvement for a local build. I see what you mean about `--local` requiring logic all over the place though. Not saying all this has to be fixed right away, just pointing out the rough edges I noticed that users are sure to run into as well.

- `doozer --local -i logging-curator5 images:rebase`

1. requires `-m` even though who cares?
2. requires `--version` although again, who cares?
3. if you don't supply `--version` then you get a TypeError stacktrace trying to match the version regex against None, instead of a helpful error

- `doozer --local -i logging-curator5 images:build`
1. requires `--repo-type` ... might be nice to explain why, or what the options mean, or default to unsigned if not supplied; something more helpful than "No repos specified. --repo-type or --repo is required."
2. also specifying `--repo` alone gets you an error `-- unable to copy transient context into container: stat .oit/None.repo: no such file or directory`
3. when imagebuilder fails, the error `Unable to start imagebuilder` is misleading -- imagebuilder may start fine, but the build itself failed. should be something like `error while running imagebuilder`
4. the experience does kinda suck without realtime output :( yum install takes forever... every time you run. output would be nice i guess, at least in the single-build case. imagebuilder now seems to send everything to stdout the way it should so maybe this would not be so bad. low priority though.
5. with unspecified release, it's put in image tag as "?" so get error:  `Committing changes to openshift3/ose-logging-curator5:v3.11.59-? ...   ; err=unable to commit build container: API error (500): {"message":"invalid tag format"}`
6. with version and release of something already built, you get `Image already built` instead of testing the build
7. maybe skip `Image build metrics:` since there won't be any brew builds?
8. images can actually build in OSBS with imagebuilder or docker. each has its place. it would be best if the local build could switch to use imagebuilder or `docker build` according to `{default_,}image_build_method`

Thanks @adammhaile

sosiouxme:
I wonder if this might work better as images:local-build? I'm not sure shoehorning the local logic alongside the brew use case is the safest approach.

adammhaile:
@sosiouxme 
I had the same `local-build` thought but rebase/update-dockerfile also needs to be "local". And others will need the same treatment eventually. So I felt it deserved more of a global option. This also means that you can add `local: true` to your `~/.config/doozer/settings.yaml` and it will always build local. Which I feel would be handy for some (though then I should give the ability to pass `--no-local` or something to temp turn it off).

rest of answers numbered...

1. Agreed, that should be made optional for local. But I knew there would be debate on how to handle the local aspect. So I held off. Honestly, that would be a tick in the column *for* having a separate verb (as -m is required currently)

2. Same answer as #1

3. Yeah, I saw the same thing. That just needs to be fixed in general and have that option be required or update the logic to make that mean "don't change the version in the Dockerfile"

Damnit, your numbers started over :P

1. `--repo-type` will go away very soon. But for now I think default it to unsigned makes sense.

2. Well... that's just a bug :P That click option should require a parameter after `--repo`

3.  Fixed.

4. Yeah, I just need to make a `cmd_realtime` or something in `exectools` that does this. Easy to do. But would probably have to only run that when I detect it's a single image build.

5. Oops. Hmm... what do you suggest? Just always set to `1` maybe if unspecified?

6. Testing the build? Is imagebuilder supposed to do that?

7. Yeah... that needs to come out. Working on it.

8. Hmmm... interesting. I'll see what I can manage.

sosiouxme:
On Thu, Dec 20, 2018 at 10:18 AM Adam Haile <notifications@github.com>
wrote:

>
>    1.
>
>     That click option should require a parameter after --repo
>
>
It does, but the build is still looking for a local repofile. I guess
--repo doesn't really jive with --local anyway. But the errmsg suggests it!


>
>    1. Oops. Hmm... what do you suggest? Just always set to 1 maybe if
>    unspecified?
>
>
Probably the best approach.

>
>    1.
>
>    Testing the build? Is imagebuilder supposed to do that?
>
>
No, I mean because there was an existing build w/ same NVR in brew, the
local build didn't run at all.


>

adammhaile:
Ah... good call Luke. It's on my list to tackle after the break :)

adammhaile:
@sosiouxme 
- Regarding `-m`, I have made it optional when `--local` is specified
- `--version` now works correctly if not provided and is not required at all for `--local`
- `--repo-type` now defaults to `unsigned`
- Fixed `--repo` issue
- Fixed `unable to commit build container` tag error
- Metrics are now skipped

Working on concurrency now...

sosiouxme:
I used this for some testing this afternoon and it was 👍
Need to take a gander at the code again...

doozer
```diff
@@ -79,6 +103,8 @@ def print_version(ctx, param, value):
               help='If a base image is not included, lookup latest FROM tag for parent. Implies --ignore-missing-base')
 @click.option("--quiet", "-q", default=False, is_flag=True, help="Suppress non-critical output")
 @click.option('--debug', default=False, is_flag=True, help='Show debug output on console.')
+@click.option("--source", metavar="ALIAS PATH", nargs=2, multiple=True,
+              help="Associate a path with a given source alias.  [multiple]")
```
sosiouxme: ... isn't this a duplicate of the next 2 lines?

adammhaile: Oops... Yeah. Will remove.

doozer
```diff
@@ -229,6 +253,11 @@ def images_update_dockerfile(runtime, stream, version, release, repo_type, messa
     - If '+', the current release will be bumped.
     - Else, the literal value will be set in the Dockerfile.
     """
+
+    if runtime.local:
+        yellow_print('images:update-dockerfile is not valid when using --local, use images:rebase instead')
```
tbielawa: I love that more and more you're using color coding for informational messages now. That extra lil' bit of context is so helpful for at-a-glance checking

doozer
```diff
@@ -707,19 +763,23 @@ def images_build_image(runtime, odcs, repo_type, repo, push_to_defaults, push_to
...
+    if not runtime.local:
+        threads = None
+
     results = runtime.parallel_exec(
         lambda (dgr, terminate_event): dgr.build_container(
             odcs, repo_type, repo, push_to_defaults, additional_registries=push_to,
-            terminate_event=terminate_event, scratch=scratch),
-        items)
+            terminate_event=terminate_event, scratch=scratch, realtime=(threads == 1)),
+        items, n_threads=threads)
```
tbielawa: An idea for _the future_ would be to consider using the hacky progress bar I created for parallel exec stuff. Akin to this which still remains in elliott

https://github.com/openshift/elliott/blob/master/elliott#L478-L480

doozer
```diff
@@ -707,19 +763,23 @@ def images_build_image(runtime, odcs, repo_type, repo, push_to_defaults, push_to
...
-        print_build_metrics(runtime)
-    except:
-        # Never kill a build because of bad logic in metrics
-        traceback.print_exc()
-        runtime.logger.error("Error trying to show build metrics")
+    if not runtime.local:  # not needed for local builds
+        try:
+            print_build_metrics(runtime)
+        except:
+            # Never kill a build because of bad logic in metrics
```
tbielawa: :+1:

doozer
```diff
@@ -829,6 +889,50 @@ def images_scan_for_cves(runtime):
...
 
+@cli.command("images:show-tree", short_help="Display the image relationship tree")
+@click.option(
+    "--imagename", default=False, is_flag=True,
+    help="Use the image name instead of the dist-git name")
+@click.option(
+    "--yml", default=False, is_flag=True,
+    help="Ouput to yaml formatted text, otherwise generate a tree view")
+@pass_runtime
+def images_print(runtime, imagename, yml):
```
tbielawa: Yo dawg, I noticed this doesn't have a docstring, so it won't have a cool `--help` output. It's not _like super important or whatever_, but it's inconsistent.

doozer
```diff
@@ -829,6 +889,50 @@ def images_scan_for_cves(runtime):
...
+
+    print_colors = ['green', 'cyan', 'blue', 'yellow', 'magenta']
+
+    def name(image):
+        return image.image_name if imagename else image.distgit_key
+
+    if yml:
+        color_print(yaml.safe_dump(runtime.image_tree, indent=2, default_flow_style=False), 'cyan')
+    else:
+        def print_branch(image, indent=0):
```
tbielawa: I assume you have a Really Good Reason™ for using two nested functions in this action?

doozer
```diff
@@ -829,6 +889,50 @@ def images_scan_for_cves(runtime):
...
+                    tree_char = '├─'
+                tree_char += ('┐' if len(child.children) else ' ')
+
+                line = '{} {} {}'.format(('  ' * indent), tree_char, name(child))
+                color_print(line, print_colors[(indent + 1) % len(print_colors)])
+                print_branch(child, indent + 1)
+
+        for image in images:
+            # print(image.name)
+            # print(image.parent)
```
tbielawa: Still need these?

doozer
```diff
@@ -1051,8 +1155,7 @@ def cleanup(runtime):
 
     runtime.logger.info('Clearing out {}'.format(runtime.working_dir))
     if os.path.isdir(runtime.working_dir):
-        shutil.rmtree(runtime.working_dir, ignore_errors=True)
-        os.makedirs(runtime.working_dir)  # rmtree deletes the directory itself. recreate
+        exectools.cmd_assert('rm -rf {}/*'.format(runtime.working_dir))
```
tbielawa: I know this already existed and you just re-wrote how it works, but since I'm just now seeing it, is there any assurance built in that this will never accidentally run somewhere like `/home/anybody` or just plain old `/`?

doozerlib/__init__.py
```diff
@@ -1,6 +1,5 @@
 from .runtime import Runtime
 from .pushd import Dir
-from .errata import Erratum
```
tbielawa: :+1:

doozerlib/exectools.py
```diff
@@ -83,7 +87,7 @@ def cmd_assert(cmd, retries=1, pollrate=60, on_retry=None, set_env=None):
     return stdout, stderr
 
 
-def cmd_gather(cmd, set_env=None):
+def cmd_gather(cmd, set_env=None, realtime=False):
```
tbielawa: No docstring for this boolean kwarg. A little exposition in the docstring would be nice.

doozerlib/exectools.py
```diff
@@ -118,9 +122,45 @@ def cmd_gather(cmd, set_env=None):
...
-    rc = proc.returncode
+
+    if not realtime:
+        out, err = proc.communicate()
+        rc = proc.returncode
+    else:
+        out = ""
+        err = ""
+
+        # Many thanks to http://eyalarubas.com/python-subproc-nonblock.html
```
tbielawa: I <3 it when people give proper credit to external resources, like you did here. :hugs:

doozer
```diff
@@ -87,6 +111,7 @@ def print_version(ctx, param, value):
               help='Process Dockerfiles in ODCS mode. HACK for the time being.')
 @click.option('--load-disabled', default=False, is_flag=True,
               help='Treat disabled images/rpms as if they were enabled')
+@click.option('--local/--osbs', default=False, is_flag=True, help='--local to run in local-only mode, --osbs to run on build cluster (default)')
```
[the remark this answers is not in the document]
adammhaile: Ok, so the reason for this was that (because of how the settings.yaml stuff works) you could set `local: true` in your settings and it would *always* run in local mode. So `--osbs` puts you back in regular mode. It was just easier to do that as a single option in click. We do that already with `--push/--no-push` which I realize makes more sense... but `--local/--no-local` is dumb.

doozerlib/kinit.py
```diff
@@ -0,0 +1,12 @@
+from exectools import cmd_gather
+from exceptions import DoozerFatalError
+
+
+def check_status():
+    rc, out, err = cmd_gather("klist")
+    if rc == 0:
+        if "No credentials cache found" in out:
+            raise DoozerFatalError('Cannot continue because you have no kerberos ticket.')
```
tbielawa: Seems a little short for a new file, but I think I get the idea. It didn't really fit in anywhere else. In the future do you think there's anywhere else this could logically live?

doozerlib/runtime.py
```diff
@@ -29,8 +29,8 @@
 from multiprocessing import Lock
 from repos import Repos
 import brew
-import constants
 from exceptions import DoozerFatalError
+import kinit
```
tbielawa: If the one `kinit` function in that new module is only used here, is there a chance you could move it here **in the future**? Less module namespace sprawl, localization of only-used-once stuff and all that jazz.

doozerlib/runtime.py
```diff
@@ -344,6 +348,32 @@ def filter_disabled(n, d):
                 if not self.image_map:
                     self.logger.warning("No image metadata directories found for given options within: {}".format(self.group_dir))
 
+                for image in self.image_map.itervalues():
+                    image.resolve_parent()
+
+                self.image_tree = {}
+                image_lists = {0: []}
+
+                def add_child_branch(child, branch, level=1):
```
tbielawa: Another time I'm not 100% sure why this has a be a nested function. Is that simply your preference for local stuff you don't think needs to live in the top level namespace for modules? I think if you expanded that argument list you could include a reference to the `image_lists` object.

doozerlib/util.py
```diff
@@ -38,3 +38,8 @@ def yellow_print(msg):
 def cprint(msg):
     """Wrapper for click.echo"""
     click.echo(msg)
+
+
+def color_print(msg, color='white', nl=True):
+    """Print out a message in given color"""
+    click.secho(msg, nl=nl, bold=False, fg=color)
```
tbielawa: At first this seems like a restrictive wrapper around `click.echo`. I looked above and saw you use it in the image tree printing stuff. This function makes those tree printing operations take fewer LOC or something?

adammhaile:
@sosiouxme hmm... I never tried that. I'll see what I can do. Probably easy fix.

doozer
adammhaile: :D   Note that I added `color_print()` which lets you give it other ANSI color names (I used this in `images:show-tree`

doozer
adammhaile: OOoo nice. I was unaware of that. Definitely.

doozer
adammhaile: Will add. And that function is named wrong. Copy/Paste error... fixing.

doozer
adammhaile: Because I do that frequently and inner functions is a common python thing (in my experience). There is no use for them outside of the scope of show-tree so no reason to put them anywhere but inside show-tree. Plus it keeps the code DRY.
https://realpython.com/inner-functions-what-are-they-good-for/

doozer
adammhaile: Nope. Gone.

doozer
adammhaile: No... and actually, this works on my laptop but not on buildvm. I'm thinking of just removing it entirely. I have no idea why it fails to do anything on buildvm, but it's not super useful anyways.

doozerlib/exectools.py
adammhaile: Done.

doozerlib/exectools.py
adammhaile: :D If I didn't I wouldn't remember where I got it from.

doozerlib/kinit.py
adammhaile: oh... this is should be removed. Doing now. I realized that it only fails if you have no klist at all, but if you have any it succeeds even if you don't have the right one. It was an experiment.

doozerlib/runtime.py
adammhaile: I removed the whole thing. Didn't really work as desired anyways.

doozerlib/runtime.py
adammhaile: It's a personal style thing for stuff that's recursive. I don't believe a function should exist in any scope it's not used.

doozerlib/util.py
adammhaile: It's because `nl` and `bold` would be the same every time. It's now less LOC but it's less typing redundant stuff. If something is always the same it belongs buried in a function so there's only one place to change it ;) So yes, it's a restrictive wrapper.

doozerlib/runtime.py
```diff
@@ -525,14 +557,14 @@ def resolve_image(self, distgit_name, required=True):
         if distgit_name not in self.image_map:
             if not required:
                 return None
-            raise IOError("Unable to find image metadata in group / included images: %s" % distgit_name)
+            raise DoozerFatalError("Unable to find image metadata in group / included images: %s" % distgit_name)
```
sosiouxme: hooray for converting nondescript IOError to real exception!

doozerlib/distgit.py
```diff
@@ -607,16 +627,20 @@ def build_container(
             # Status defaults to failure until explicitly set by success. This handles raised exceptions.
         }
 
-        target_tag = "-".join((self.org_version, release))
+        if self.runtime.local and release is '?':
+            release_val = ''
+        else:
+            release_val = '-{}'.format(release)
+        target_tag = "{}{}".format(self.org_version, release_val)
```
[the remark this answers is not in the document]
adammhaile: ack, done.

doozerlib/distgit.py
```diff
@@ -607,16 +627,20 @@ def build_container(
...
 
         try:
-            if not scratch and self.org_release is not None \
+            if not self.runtime.local and not scratch and self.org_release is not None \
                     and self.metadata.tag_exists(target_tag):
                 self.logger.info("Image already built for: {}".format(target_image))
             else:
                 # If this image is FROM another group member, we need to wait on that group member
-                # Use .get('from',None) since from is a reserved word.
+                # Use .get('from',None) since from is a reservbuild_conted word.
```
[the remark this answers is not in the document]
adammhaile: fixed

doozerlib/distgit.py
```diff
@@ -691,12 +719,58 @@ def wait(n):
...
 
+    def _build_container_local(self, target_image, repo_type, realtime=False):
+        """
+        The part of `build_container` which actually starts the build,
+        separated for clarity. Local build version.
+        """
+
+        if self.image_build_method == 'imagebuilder':
+            builder = 'imagebuilder -mount '
+            z_add = ''
```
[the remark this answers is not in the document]
adammhaile: yep, removed

doozerlib/distgit.py
```diff
@@ -943,17 +994,24 @@ def update_distgit_dir(self, version, release):
...
-            with open('additional-tags', 'w') as at:
-                at.write("%s\n" % uuid_tag)  # The uuid which we ensure we get the right FROM tag
-                # at.write("%s\n" % version)  # Removed for https://projects.engineering.redhat.com/browse/OSBS-5638?focusedCommentId=837662&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-837662
-                vsplit = version.split(".")
-                if len(vsplit) > 1:
-                    at.write("%s.%s\n" % (vsplit[0], vsplit[1]))  # e.g. "v3.7.0" -> "v3.7"
+            if not self.runtime.local:
+                with open('additional-tags', 'w') as at:
+                    at.write("%s\n" % uuid_tag)  # The uuid which we ensure we get the right FROM tag
+                    # at.write("%s\n" % version)  # Removed for https://projects.engineering.redhat.com/browse/OSBS-5638?focusedCommentId=837662&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-837662
```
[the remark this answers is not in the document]
adammhaile: done.

doozerlib/runtime.py
adammhaile: :D

adammhaile:
@sosiouxme @tbielawa nope, good to go :) Merging.

PR #20, opened 2019-01-02
Only add images that exist, sorted the ImageStream

tbielawa:
Uses `skopeo inspect` to check if a given input image actually exists before adding it to the sync list.

The ImageStream `tags` list is now sorted.

### DIFF
--- a/doozer
+++ b/doozer
@@ -1267,6 +1267,7 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
     src_dest_items = []
 
     for image in images:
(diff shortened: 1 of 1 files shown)


tbielawa:
This is dirty right now. Needs to be cleaned up. But this is the beginning of getting things fixed for issues that @smarterclayton and I ran into before break.

tbielawa:
I'm going to merge this because I need it on buildvm to test back and forth with Clayton right now. It will obviously get improved real soon.

PR #27, opened 2019-01-16
Clone sources at depth 1

adammhaile:
@tbielawa @sosiouxme

### DIFF
--- a/doozerlib/runtime.py
+++ b/doozerlib/runtime.py
@@ -644,56 +644,62 @@ class Runtime(object):
 
         url = source_details["url"]
         branches = source_details['branch']
-        self.logger.info("Cloning source '%s' from %s as specified by group into: %s" % (alias, url, source_dir))
-        exectools.cmd_assert(
-            cmd=["git", "clone", url, source_dir],
-            retries=3,
-            on_retry=["rm", "-rf", source_dir],
-        )
+
         stage_branch = branches.get('stage', None)
         fallback_branch = branches.get("fallback", None)
         found = False
-        with Dir(source_dir):
-            if self.stage and stage_branch:
-                self.logger.info('Normal branch overridden by --stage option, using "{}"'.format(stage_branch))
-                branch = stage_branch
-            else:
-                branch = branches["target"]
-            self.logger.info("Attempting to checkout source '%s' branch %s in: %s" % (alias, branch, source_dir))
 
-            if branch != "master":
-                rc, out, err = exectools.cmd_gather(["git", "checkout", "-b", branch, "origin/%s" % branch])
-            else:
-                rc = 0
+        if self.stage and stage_branch:
+            self.logger.info('Normal branch overridden by --stage option, using "{}"'.format(stage_branch))
+            branch = stage_branch
+        else:
+            branch = branches["target"]
 
-            if rc == 0:
-                found = True
+        clone_branch = None
+
+        branch_check = 'git ls-remote --heads {} {}'
+        for _ in range(3):
+            self.logger.info('Checking if target branch {} exists in {}'.format(branch, url))
+            rc, out, err = exectools.cmd_gather(branch_check.format(url, branch))
+            if rc:
+                self.logger.error('Unable to check if target branch {} exists: {}'.format(branch, err))
+                continue  # retry
+            if out.strip():  # any result means the branch is found
+                clone_branch = branch
+                break
+            elif not fallback_branch:
+                raise DoozerFatalError('Requested target branch {} does not exist and no fallback provided'.format(branch))
             else:
                 if self.stage and stage_branch:
                     raise DoozerFatalError('--stage option specified and no stage branch named "{}" exists for {}|{}'.format(stage_branch, alias, url))
-                elif fallback_branch is not None:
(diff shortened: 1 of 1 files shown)


doozerlib/runtime.py
```diff
@@ -644,56 +644,44 @@ def resolve_source(self, parent, meta):
...
+            branch = branches["target"]
 
-            if rc == 0:
-                found = True
-            else:
+        base_clone = "git clone -b {} --single-branch {} --depth 1 {}"
+        clone_target = base_clone.format(branch, url, source_dir)
+        clone_fallback = base_clone.format(fallback_branch, url, source_dir)
+        retries = 3
+        while retries:
```
sosiouxme: could we get rid of the iterator and just do this instead:
```suggestion
        while _ in range(3):
```

doozerlib/runtime.py
```diff
@@ -644,56 +644,44 @@ def resolve_source(self, parent, meta):
...
+            try:
+                self.logger.info("Attempting to checkout source '%s' branch %s in: %s" % (alias, branch, source_dir))
+                exectools.cmd_assert(clone_target)
+            except IOError:
                 if self.stage and stage_branch:
                     raise DoozerFatalError('--stage option specified and no stage branch named "{}" exists for {}|{}'.format(stage_branch, alias, url))
-                elif fallback_branch is not None:
```
sosiouxme: I think this should still check whether there's a fallback branch before trying to use it.

doozerlib/runtime.py
```diff
@@ -644,56 +644,44 @@ def resolve_source(self, parent, meta):
...
+                try:
```
sosiouxme: Also it bothers me a bit to go right from an attempted clone to the fallback, and without cleaning up. We don't know why the first one failed... maybe it was part-way cloned and then there was a network bobble. We would probably want to rm the directory and try again. We would probably not want to try the backup without checking to see if the initial branch even existed. Maybe a git remote-ls first to decide what can be cloned, then you just have a single thing to clone (and retry)?

adammhaile:
Ok @sosiouxme and @tbielawa - I've updated it to verify the existence of branches first with `git ls-remote` (with retries even), and then once it's found the one branch it should check out it will only attempt to checkout that one branch. Take a peep and lemme know what you think.

doozerlib/runtime.py
```diff
@@ -644,56 +644,58 @@ def resolve_source(self, parent, meta):
...
-            if found:
-                # Store so that the next attempt to resolve the source hits the map
-                self.register_source_alias(alias, source_dir)
-                return source_dir
-            else:
-                raise DoozerFatalError("Error checking out target branch of source '%s' in: %s" % (alias, source_dir))
+        for _ in range(3):  # attempt checkout 3 times
+            try:
+                self.logger.info("Attempting to checkout source '%s' branch %s in: %s" % (url, clone_branch, source_dir))
+                exectools.cmd_assert("git clone -b {} --single-branch {} --depth 1 {}".format(clone_branch, url, source_dir))
```
sosiouxme: Couldn't this reduce to:
```suggestion
                exectools.cmd_assert(
                  "git clone -b {} --single-branch {} --depth 1 {}".format(clone_branch, url, source_dir),
                  retries=3,
                  on_retry=["rm", "-rf", source_dir],
                )
```
Simpler to skip the for loop, try, and setting found; just let this fail as needed.

doozerlib/runtime.py
```diff
@@ -644,56 +644,58 @@ def resolve_source(self, parent, meta):
...
-                        rc2 = 0
 
-                    if rc2 == 0:
-                        found = True
-                    else:
-                        self.logger.error("Failed checking out fallback-branch %s: %s" % (branch, err))
-                else:
-                    self.logger.error("Failed checking out branch %s: %s" % (branch, err))
+                self.logger.info('Target branch does not exist in {}, checking fallback branch {}'.format(url, fallback_branch))
+                rc, out, err = exectools.cmd_gather(branch_check.format(url, fallback_branch))
```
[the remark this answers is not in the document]
adammhaile: I was going to but @tbielawa twitches when I do internal functions :P

doozerlib/runtime.py
adammhaile: ack, 100% right... and I forgot the `rm -rf` between attempts. Good call.

doozerlib/runtime.py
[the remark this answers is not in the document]
adammhaile: also I wanted separate messages for each... another reason I left it separate.

doozerlib/runtime.py
```diff
@@ -644,56 +644,58 @@ def resolve_source(self, parent, meta):
...
+            rc, out, err = exectools.cmd_gather(branch_check.format(url, branch))
+            if rc:
+                self.logger.error('Unable to check if target branch {} exists: {}'.format(branch, err))
+                continue  # retry
+            if out.strip():  # any result means the branch is found
+                clone_branch = branch
+                break
             else:
                 if self.stage and stage_branch:
                     raise DoozerFatalError('--stage option specified and no stage branch named "{}" exists for {}|{}'.format(stage_branch, alias, url))
```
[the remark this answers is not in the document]
adammhaile: yep... added a check for that.

adammhaile:
@sosiouxme addressed concerns.

PR #29, opened 2019-01-29
add ability to scan sources for updates

sosiouxme:
This allows us to determine when upstream sources have a different commit than the one we recorded in distgit, without cloning either. It works on both RPMs and images, and can spit out yaml for our jenkins job to grok. Examples:

```
🐚 doozer -r '' config:scan-sources 
2019-01-29 23:06:33,907 INFO Using branch from group.yml: rhaos-4.0-rhel-7
2019-01-29 23:06:34,233 INFO Checking if target branch release-4.0 exists in git@github.com:openshift/service-catalog.git
2019-01-29 23:06:34,233 INFO Checking if target branch release-4.0 exists in git@github.com:openshift/service-idler.git
2019-01-29 23:06:34,233 INFO Checking if target branch openshift-4.0 exists in git@github.com:openshift/autoheal.git
2019-01-29 23:06:34,740 INFO Target branch does not exist in git@github.com:openshift/autoheal.git, checking fallback branch master
2019-01-29 23:06:34,740 INFO Target branch does not exist in git@github.com:openshift/service-idler.git, checking fallback branch master
2019-01-29 23:06:34,741 INFO Checking if target branch master exists in git@github.com:openshift/autoheal.git
2019-01-29 23:06:34,742 INFO Target branch does not exist in git@github.com:openshift/service-catalog.git, checking fallback branch master
2019-01-29 23:06:34,742 INFO Checking if target branch master exists in git@github.com:openshift/service-idler.git
2019-01-29 23:06:34,744 INFO Checking if target branch master exists in git@github.com:openshift/service-catalog.git
2019-01-29 23:06:35,742 ERROR [rpms/openshift-enterprise-service-idler] No OS_GIT_COMMIT found in specfile 'service-idler.spec'
RPMS:
  openshift-enterprise-service-idler is changed
  openshift-enterprise-service-catalog is the same
  openshift-enterprise-autoheal is the same

🐚 doozer -i openshift-enterprise-console,ose-installer,ose-etcd,logging-fluentd -r '' config:scan-sources --yaml
2019-01-29 23:07:14,528 INFO Using branch from group.yml: rhaos-4.0-rhel-7
[...]
---
images:
- {changed: true, name: openshift-enterprise-console}
- {changed: false, name: ose-etcd}
- {changed: true, name: ose-installer}
- {changed: false, name: logging-fluentd}
rpms:
- {changed: true, name: openshift-enterprise-service-idler}
- {changed: false, name: openshift-enterprise-service-catalog}
- {changed: false, name: openshift-enterprise-autoheal}
```

Some notes:
* Does not pay any attention at all to local checkouts of sources or distgits.
* I considered just naming it something top-level, but it works somewhat similarly to other `config:` commands. Open to suggestions here.
* I thought it might be helpful to have the atomic-openshift RPM definition here, but disabled, so that we could run `doozer -r atomic-openshift config:scan-sources` and know whether we actually needed to rebuild that. I'm not sure if it has value though since we do the merge. Maybe once the merge job is separate?...
* Not all RPM specfiles have the necessary bits to capture the source commit. This is why the ERROR above and results in the source always being counted as "changed".

### DIFF
--- a/doozer
+++ b/doozer
@@ -392,6 +392,36 @@ def images_verify(runtime, image, no_pull, repo_type, **kwargs):
     exit(failed_images_count)
 
 
+@cli.command("config:scan-sources", short_help="Determine if source repo differs from distgit content")
+@click.option("--yaml", "as_yaml", default=False, is_flag=True, help='Print results in a yaml block')
+@pass_runtime
+def config_scan_source_changes(runtime, as_yaml):
+    """
+    Determine if source repo differs from distgit content, according to source git hash.
+    May be used for images and/or rpms.
+    Print a list of configs that were scanned and whether the source differs from the distgit.
+    """
+    _fix_runtime_mode(runtime)
+    runtime.initialize(**CONFIG_RUNTIME_OPTS)
+    results = dict(rpms=[], images=[])
+    for meta, matches in runtime.scan_distgit_sources():
+        results['rpms' if meta.meta_type == 'rpm' else 'images'].append(
+            dict(name=meta.distgit_key, changed=not matches)
+        )
+
+    if as_yaml:
+        print '---'
+        print yaml.safe_dump(results, indent=4)
+        return
+
+    for kind, items in results.items():
+        if not items:
+            continue
+        print kind.upper() + ":"
+        for item in items:
+            print '  {} is {}'.format(item['name'], 'changed' if item['changed'] else 'the same')
+
+
 @cli.command("images:rebase", short_help="Refresh a group's distgit content from source content.")
 @click.option("--stream", metavar="ALIAS REPO/NAME:TAG", nargs=2, multiple=True,
               help="Associate an image name with a given stream alias.  [multiple]")
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -20,8 +20,8 @@ import exectools
 from pushd import Dir
 from brew import watch_task, check_rpm_buildroot
 from model import Model, Missing
-from . exceptions import DoozerFatalError
-from . util import yellow_print
+from doozerlib.exceptions import DoozerFatalError
+from doozerlib.util import yellow_print
 
 OIT_COMMENT_PREFIX = '#oit##'
 OIT_BEGIN = '##OIT_BEGIN'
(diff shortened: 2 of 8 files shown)


sosiouxme:
Actually, not quite ready for real review. Finding some issues to fix.

adammhaile:
Out till next week :( I'm really interested in this... Can it wait to be
put in until I can review?

sosiouxme:
Sure, although I'll be out next week :)
BTW, issues fixed.

tbielawa:
/lgtm

I just tried this out and it is awesome. That was much faster than I expected it to be.

Only thing I would change in the future is how the YAML is output. For example, adding a `-o` type option to say "write output to <FILE>". And maybe grouping the changed vs. not changed stuff a little differently. Really depends on how we plan on using this.

I could see this being one step of several in a job. In the sync job I run doozer to generate input files for other commands. This could generate input for doozer itself or have the image build command pull from this command if it can pass the generated datas to it.

Or, we could have the option of invoking it as part of the custom build job! Parameter to custom build job `only build changed?`.

So many possibilities! I'm excited.

tbielawa:
also, :+1: for adding unit tests.

Some brain dumping while I'm in here:

I would like to see more docstring examples for `config:scan-sources` in the next round. The way the global `-r` handling works isn't exactly intuitive and seems to conflict with how `config:scan-sources` works. `-r` says it is *all by default*, so if I omit the `-r ''` then I get back `config:scan-sources` with no rpms listed.

But when I run `config:scan-sources` with `-r ''` I get back several RPMs in the list. That seems to conflict with the documented "all by default" behavior.

Other optional output selectors for the command would be useful, i.e., only return items which have changed.

And, honestly, I think you could drop the basic plain text output and make yaml the default. It's presenting the same information and adding extra presentation layer fluff to have to maintain. Switching `--yaml` to `--output, -o` would be value, imho.

sosiouxme:
@tbielawa 

> Or, we could have the option of invoking it as part of the custom build job! Parameter to custom build job `only build changed?`.

Or maybe a custom job like https://github.com/openshift/aos-cd-jobs/pull/1691 ?

tbielawa:
> @tbielawa
> 
> > Or, we could have the option of invoking it as part of the custom build job! Parameter to custom build job `only build changed?`.
> 
> Or maybe a custom job like [openshift/aos-cd-jobs#1691](https://github.com/openshift/aos-cd-jobs/pull/1691) ?

Yes. Exactly like that!

PR #31, opened 2019-02-07
Adding reposync command

adammhaile:
@tbielawa @jupierce This is one half of the task of syncing repos. It will sync the entire repo set used by any OCP version to a given local directory. From there the rest of the tasks should be all jenkins-side to sync that up to `use-mirror`.

Note:

- I used the `-n` option which only syncs the newest version of a package in the repo
- I used `--delete` so that it should clear out old packages when resynced
- The above are meant to be used such that the local sync dir is never deleted between syncs to speed up subsequent sync tasks.
- Currently, for 4.0, this results in 24GB and 19,000 packages.

### DIFF
--- a/doozer
+++ b/doozer
@@ -21,6 +21,7 @@ import yaml
 import sys
 import subprocess
 import urllib
+import tempfile
 import traceback
 import koji
 from numbers import Number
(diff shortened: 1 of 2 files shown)


doozer
```diff
@@ -1483,13 +1484,80 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
...
+    """
+    runtime.initialize(clone_distgits=False)
+    repos = runtime.repos
+
+    yum_conf = """
+[main]
+cachedir=/tmp/cache/yum/$basearch/$releasever
+keepcache=0
+debuglevel=2
+logfile=/tmp/yum.log
```
jupierce: better to put in the workspace?

doozer
```diff
@@ -1483,13 +1484,80 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
...
+exactarch=1
+obsoletes=1
+gpgcheck=1
```
jupierce: Consider `gpgcheck=0` to avoid surprises. i.e. sync what is there, regardless of whether it is signed.

doozer
```diff
@@ -1483,13 +1484,80 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
...
+
+        $ doozer --group=openshift-4.0 beta:reposync -o yum.conf --repo-type unsigned
+
+    """
+    runtime.initialize(clone_distgits=False)
+    repos = runtime.repos
+
+    yum_conf = """
+[main]
+cachedir=/tmp/cache/yum/$basearch/$releasever
```
jupierce: better to put in the workspace?

doozer
adammhaile: Fair. Will change.

PR #36, opened 2019-02-20
exectools.cmd_gather should not busy-wait on output

sosiouxme:


tbielawa:
/lgtm
Sounds good to me. But why?

adammhaile:
/lgtm
/approve
@tbielawa because a no-wait loop will peg the CPU. Telling it to sleep even barely will let the CPU context switch. In C I used to be able to do `sleep(0)` but in Python you need to give it a non-zero number.

PR #37, opened 2019-02-20
Adding @tbielawa oc tweak for image push

adammhaile:


tbielawa:
I got myself worried about max command line length since we mirror a BUNCH of stuff sometimes. I did some quick peeping, we're probably good on buildvm, but I don't know the equivalent answer for upshift.

Number of characters in the last mirror input file: `21058 oc_mirror_input`

Max length for command line args on buildvm:
```
$ ssh buildvm.openshift.eng.bos.redhat.com getconf ARG_MAX                                                                                                
2097152
```

If we wanted to keep mirroring inputs as build artifacts (which is an appealing idea) we would simply need to write out the `SRC=DEST` line to a file and provide it to `oc` as an option.

adammhaile:
@tbielawa I'll do it whatever way makes you happy. Personally, just moving it over to using a temp file now is the prudent choice.

tbielawa:
> @tbielawa I'll do it whatever way makes you happy. Personally, just moving it over to using a temp file now is the prudent choice.

Yeah! But not a temp file, let's make sure it gets archived with the other build artifacts :-) I actually have uses for those from the sync job sometimes.

`oc image mirror --filename=/path/to/thing` is the option you'll be looking for.

adammhaile:
It shall be done

tbielawa:
/lgtm

I did a `--dry-run` test and it seemed fantastic. Let's take this thing for a ride!

PR #38, opened 2019-02-21
Fixing state save

adammhaile:


adammhaile:
@tbielawa needed urgently :P

PR #41, opened 2019-02-28
Adding dist-git-ignore option

adammhaile:


doozerlib/distgit.py
```diff
@@ -40,6 +40,8 @@
 ---
 """
 
+BASE_IGNORE = [".git", ".oit", ".gitignore", "additional-tags"]
```
tbielawa: Comment for this plz: context as to why this variable exists

adammhaile:
*@tbielawa *comment added

doozerlib/distgit.py
tbielawa: yes

sosiouxme: although i might say that we're "ignoring" them in the sense that they don't get cleared out of distgit on rebase, not a rebase ignores source content.

adammhaile:
Sam had a similar content since ignoring implies that we also won't
overwrite... which would also be handy.

PR #43, opened 2019-03-01
add config:print --yaml

sosiouxme:
Existing output format is unchanged.
This adds an option to produce yaml that a groovy script could parse.

### DIFF
--- a/doozerlib/config.py
+++ b/doozerlib/config.py
@@ -76,39 +76,45 @@ class MetaDataConfig(object):
         for rpm in self.runtime.rpm_metas():
             self.update_meta(rpm, key, val)
 
-    def config_print(self, key=None, name_only=False):
+    def config_print(self, key=None, name_only=False, as_yaml=False):
         """
         Print name, sub-key, or entire config
         """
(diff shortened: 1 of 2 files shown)


doozerlib/config.py
```diff
@@ -76,39 +76,45 @@ def update(self, key, val):
...
+        _collect_data(self.runtime.image_metas(), data["images"])
+        _collect_data(self.runtime.rpm_metas(), data["rpms"])
 
-        if image_metas:
-            print('')
-            print('********* Images *********')
-            for img in image_metas:
-                _do_print(img, key)
+        if as_yaml:
+            print yaml.safe_dump(data, default_flow_style=False)
```
adammhaile: needs ()
Though in general, I feel these should be switched to our `*_print` functions as they use click print which is safer.
But that's, I admit, out of the context of this PR as I was the one that did that in the first place :P

sosiouxme: works without the parens, but sure, consistency.
also feel free to go through and convert all these to a proper print...

PR #44, opened 2019-03-01
distgit: extract container build logs from brew logs

sosiouxme:
These will be useful for debugging, or sending to image owners so they can debug.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -895,6 +895,7 @@ class ImageDistGitRepo(DistGitRepo):
         if logs_rc != 0:
             self.logger.info("Error downloading build logs from brew for task %s: %s" % (task_id, logs_err))
         elif error is None:
+            self._extract_container_build_logs(task_id)
             # even if the build completed without error, check the logs for problems
             error = self._detect_permanent_build_failures(self.runtime.group_config.image_build_log_scanner)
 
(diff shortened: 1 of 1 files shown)


doozerlib/distgit.py
```diff
@@ -906,8 +907,37 @@ def _build_container(
...
+        """
+        logs_dir = self._logs_dir(task_id)
+        try:
+            for filename in os.listdir(logs_dir):
+                rc, output, _ = exectools.cmd_gather([
+                    "sed", "-nEe",
+                    # look in logs for either of the build plugins to start,
+                    # and look for the plugin runner to declare the plugin done to end.
+                    # strip off all the noise at the front of log lines.
+                    "/plugins.(imagebuilder|docker_api)/,/atomic_reactor.plugin / { s/^.*?- (DEBUG|INFO) - //; p }",
```
tbielawa: I highly appreciate commented regexes <3

doozerlib/distgit.py
```diff
@@ -906,8 +907,37 @@ def _build_container(
         self.logger.info("Successfully built image: {} ; {}".format(target_image, task_url))
         return True
 
-    def _logs_dir(self):
-        return "%s/%s" % (self.runtime.brew_logs_dir, self.metadata.name)
+    def _logs_dir(self, task_id=None):
+        segments = [self.runtime.brew_logs_dir, self.metadata.name]
+        if task_id:
```
tbielawa: I think `pep` would tell you `if task_is is not None`, but whatever.

tbielawa:
/lgtm

This will be great!

doozerlib/distgit.py
adammhaile: @tbielawa no, this is valid as long as you really are just looking at the "truthiness" of the variable. 
I agree that, for consistency if should be `if task_id is not None` because that's the default param value given. If you are OK with any value where `bool(task_id) is False` then this is more concise and I'm OK with it.

sosiouxme: I went ahead and put `is not None` in there, as the intent is arguably clearer that way.

PR #48, opened 2019-03-05
Fixing push state messages

adammhaile:
@sosiouxme noted that failed pushes didn't make it into state.yaml which should be fixed by this.
Annoyingly the assert/exec/parallel_exec stuff makes it a royal pain... I'm open to thoughts on if there's a better way to do it.
Also... TIL that not all Exception objects have a `message` attribute... which is freaking dumb. That's why there's a couple `repr(ex)` calls in there. No other way to get the message out of the exceptions thrown in the `assertion.success()` call.
@tbielawa

### DIFF
--- a/doozer
+++ b/doozer
@@ -307,12 +307,16 @@ def images_update_dockerfile(runtime, stream, version, release, repo_type, messa
 
     try:
         if push:
-            runtime.push_distgits()
+            res = runtime.push_distgits()
 
-        state.record_image_finish(lstate)
+            for name, r in res:
+                if r is not True:
+                    state.record_image_fail(lstate, name, r)
     except Exception as ex:
         lstate['status'] = state.STATE_FAIL
-        raise DoozerFatalError(ex.message)
+        raise DoozerFatalError(getattr(ex, 'message', repr(ex)))
+
+    state.record_image_finish(lstate)
 
     failed = []
     for img, status in lstate['images'].iteritems():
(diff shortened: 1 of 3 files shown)


doozer
```diff
@@ -307,12 +307,16 @@ def images_update_dockerfile(runtime, stream, version, release, repo_type, messa
...
+            res = runtime.push_distgits()
 
-        state.record_image_finish(lstate)
+            for name, r in res:
+                if r is not True:
+                    state.record_image_fail(lstate, name, r)
     except Exception as ex:
         lstate['status'] = state.STATE_FAIL
-        raise DoozerFatalError(ex.message)
+        raise DoozerFatalError(repr(ex))
```
sosiouxme: This seems fine to me, but there's a way to use both if you think it's better:
https://stackoverflow.com/a/45532289
```suggestion
        raise DoozerFatalError(getattr(ex, 'message', repr(ex)))
```

doozerlib/distgit.py
```diff
@@ -981,11 +981,15 @@ def _detect_permanent_build_failures(self, scanner):
...
-            # and tags in dist-git can't be --force overwritten
-            exectools.cmd_gather(['timeout', '60', 'git', 'push', '--tags'])
+            try:
+                exectools.cmd_assert("timeout 1200 rhpkg push", retries=3)
+                # rhpkg will create but not push tags :(
+                # Not asserting this exec since this is non-fatal if a tag already exists,
+                # and tags in dist-git can't be --force overwritten
+                exectools.cmd_gather(['timeout', '60', 'git', 'push', '--tags'])
+            except IOError as e:
+                return (self.metadata, repr(e.message))
```
sosiouxme: `repr`? is `e.message` ever not a string?

adammhaile: oops... that was a typo, should've removed `.message`

sosiouxme:
/lgtm
:ship: :shipit:

PR #49, opened 2019-03-05
Always delete .gitignore for distgits

adammhaile:
@sosiouxme @joeldavis84 
as discussed

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -42,7 +42,7 @@ CONTAINER_YAML_HEADER = """
 
 # Always ignore these files/folders when rebasing into distgit
 # May be added to based on group/image config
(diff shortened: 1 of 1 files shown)


doozerlib/distgit.py
```diff
@@ -1374,6 +1374,12 @@ def _merge_source(self):
             if ent.startswith("Dockerfile."):
                 os.remove(ent)
 
+        # Delete .gitignore since it may block full sync and is not needed here
+        try:
+            os.remove("./.gitignore")
```
sosiouxme: should we also remove it from BASE_IGNORE? doesn't matter much but more tidy...

sosiouxme:
lgtm

PR #54, opened 2019-03-15
distgit.py: clone with the branch to begin with

sosiouxme:
Used to rhpkg clone and then switch branch. As far as I can see it's faster to get it right the first time than change it afterward. Don't see any need to keep it separate steps.


tbielawa:
I'd buy this for a dollar
/lgtm

PR #55, opened 2019-03-16
rpmcfg.py: roll back the temporary commit when done

sosiouxme:
Tidy up after an RPM build by removing the commit that's added for the
build. In this way the source for an alias is not polluted with commits
that can be picked up by other builds from that alias.


sosiouxme:
WIP because I realized this relates to the problem I saw with tito and ose tags. Adding another change to consider.

sosiouxme:
Gonna go with this for now to unblock builds.

PR #56, opened 2019-03-19
Define 'blacklist' in group schema

tbielawa:
Bug: https://bugzilla.redhat.com/show_bug.cgi?id=1690189

Accompanies a change in ocp-build-data.


tbielawa:
I am writing up the code to ingest this new blacklist parameter. It will arrive as an additional pull request.

tbielawa:
Not required.

PR #57, opened 2019-03-19
Blacklist sync

tbielawa:
Do not sync images that are on the blacklist.

Child PR of  https://github.com/openshift/doozer/pull/56/files

### DIFF
--- a/doozer
+++ b/doozer
@@ -1539,8 +1539,8 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
           namespace: ocp
         spec:
           tags: []
-
     """
+    runtime.exclude.extend(runtime.group_config.get('blacklist', []))
     runtime.initialize(clone_distgits=False)
     images = [i for i in runtime.image_metas()]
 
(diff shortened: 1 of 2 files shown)


doozer
```diff
@@ -1539,8 +1539,8 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
           namespace: ocp
         spec:
           tags: []
-
     """
+    runtime.exclude.extend(runtime.group_config.get('blacklist', []))
```
adammhaile: Pending name decision of course.

doozer
```diff
@@ -1552,6 +1552,7 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
     src_dest_items = []
     missing_source_items = []
     invalid_name_items = []
+    blacklisted_items = []
```
adammhaile: This doesn't get populated anywhere that I see.... what am I missing? what's the point?

tbielawa:
Not required

PR #64, opened 2019-03-22
doozer.release_mirror: init runtime so group_config exists

sosiouxme:
`group_config` doesn't exist before `initialize`.

However I'm not certain the exclusion actually works for this purpose if it's set after initialization. Check me :)

### DIFF
--- a/doozer
+++ b/doozer
@@ -1553,8 +1553,11 @@ def release_mirror(runtime, src_dest, image_stream, is_base, pattern):
           tags: []
 
     """
-    runtime.exclude.extend(runtime.group_config.get('non_release.images', []))
     runtime.initialize(clone_distgits=False)
+    non_release_images = runtime.group_config.non_release.images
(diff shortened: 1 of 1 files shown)


adammhaile:
Arg.... :( Yeah.... @tbielawa yeah, I led you slightly astray here. @sosiouxme is 100% right that group_config doesn't exist until after initialize at which time it's too late. In fact it hasn't even fetched ocp-build-data at that point :( The only solution I can think of is to add a param to `Runtime.initialize()` to the effect to `exclude_non_release=True` that is default false. Then when initializing in `release_mirror` set the value to true.... THEN, after this line -> https://github.com/openshift/doozer/blob/master/doozerlib/runtime.py#L322

do basically the same extension of exclude list from group_config that was already attempted. Far less elegant I know, but only way. @tbielawa you may now berate me for sending you down an incorrect path.

PR #66, opened 2019-03-24
important fixes for broken stuff

sosiouxme:


adammhaile:
@sosiouxme can you explain why these were needed?

sosiouxme:
@adammhaile good point, I only did that in the commit messages.

### repos.py: fix for pulp API change

The content sets query asked for a specific subfield for `notes`, and
pulp no longer returns that. So, now queries for the entire `notes`
field.

### runtime.py: remove errant print

This broke things that are trying to produce yaml output

### image.py: remove extraneous 'import container'

... because there is no `container` module.

adammhaile:
ACk, thanks @sosiouxme

PR #69, opened 2019-03-28
ART-514: print out tags in necessary RPM builds for 4.1

shiywang:


doozerlib/distgit.py
```diff
@@ -935,6 +941,16 @@ def _extract_container_build_logs(self, task_id):
...
 
+    def add_missing_pkgs(self, pkg_name):
+        """
+        add missing packages to runtime.missing_pkgs set and print it out for
+        `brew tag-build` from `rhaos-4.0-rhel-7-candidate` to `rhaos-4.1-rhel-7-candidate` manually
+        :param pkg_name: Missing packages like: No package (.*) available
+        """
+        latest = image.exec_latest_build_info("rhaos-4.0-rhel-7-candidate", pkg_name)
+        if not latest.startswith(pkg_name):
+            self.runtime.missing_pkgs.add(pkg_name)
```
adammhaile:
```suggestion
            with self.runtime.mutex:
                self.runtime.missing_pkgs.add(pkg_name)
```

Because this is threaded you need to get a lock on Runtime

doozer
```diff
@@ -851,6 +851,8 @@ def images_build_image(runtime, odcs, repo_type, repo, push_to_defaults, push_to
     failed = [name for name, r in results if not r]
     if failed:
         runtime.logger.error("\n".join(["Build/push failures:"] + sorted(failed)))
+        if len(runtime.missing_pkgs) != 0:
+            runtime.logger.error("Missing pkgs are: \n".join(str(s) for s in runtime.missing_pkgs))
```
adammhaile:
```suggestion
            runtime.logger.error("Missing pkgs are: \n{}".format("\n".join(runtime.missing_pkgs))
```
Your code was joining on `Missing pkgs are:` which it would've printed out one for each package.
Also, the set is already a set of strings, no need to iterate and convert to string,

doozerlib/image.py
```diff
@@ -281,3 +277,8 @@ def get_additional_push_names(self, additional_registries):
                 push_names.append('{}/{}/{}'.format(registry, namespace, repo_name))
 
         return push_names
+
+def exec_latest_build_info(tag, component_name):
+    rc, stdout, stderr = exectools.cmd_gather(["brew", "latest-build", tag, component_name])
+    assertion.success(rc, "Unable to search brew builds: %s" % stderr)
```
adammhaile: Please change this to a `.format()` notation. `%` is deprecated. (I realize you probably just copied it from elsewhere in here and that's ok... but I try to remove the old form anytime I come across one).

doozerlib/distgit.py
```diff
@@ -709,10 +710,15 @@ def wait(n):
                         # No need to retry if the failure will just recur
                         error = self._detect_permanent_build_failures(self.runtime.group_config.image_build_log_scanner)
                         if error is not None:
-                            raise exectools.RetryException(
-                                "Saw permanent error in build logs:\n{}\nWill not retry after {} failed attempt(s)"
-                                .format(error, n + 1)
-                            )
+                            need_tag = re.match("No package (.*) available", error)
```
sosiouxme:
```suggestion
                            for match in re.finditer("No package (.*) available", error):
                                missing_pkg = match.group(1)
```

doozerlib/distgit.py
```diff
@@ -709,10 +710,15 @@ def wait(n):
...
+                            if need_tag:
+                                missing_pkg = re.search("No package (.*) available", error).group(1)
+                                self.add_missing_pkgs(missing_pkg)
+                            else:
```
sosiouxme: don't need else -- unconditionally throw exception

doozerlib/distgit.py
```diff
@@ -935,6 +941,17 @@ def _extract_container_build_logs(self, task_id):
         except OSError as e:
             self.logger.warning("Exception while trying to extract build logs in {}: {}".format(logs_dir, e))
 
+    def add_missing_pkgs(self, pkg_name):
+        """
+        add missing packages to runtime.missing_pkgs set and print it out for
+        `brew tag-build` from `rhaos-4.0-rhel-7-candidate` to `rhaos-4.1-rhel-7-candidate` manually
+        :param pkg_name: Missing packages like: No package (.*) available
+        """
+        latest = image.exec_latest_build_info("rhaos-4.0-rhel-7-candidate", pkg_name)
```
sosiouxme: tag cannot be hardcoded but i'm not sure where to get it :)

get from group.yml branch and decrement x in 4.x

doozerlib/image.py
```diff
@@ -281,3 +277,8 @@ def get_additional_push_names(self, additional_registries):
                 push_names.append('{}/{}/{}'.format(registry, namespace, repo_name))
 
         return push_names
+
+def exec_latest_build_info(tag, component_name):
```
sosiouxme: unit test this

shiywang:
@sosiouxme updated, PTAL

doozerlib/distgit.py
```diff
@@ -935,6 +937,17 @@ def _extract_container_build_logs(self, task_id):
         except OSError as e:
             self.logger.warning("Exception while trying to extract build logs in {}: {}".format(logs_dir, e))
 
+    def _add_missing_pkgs(self, pkg_name):
+        """
+        add missing packages to runtime.missing_pkgs set and print it out for
+        `brew tag-build` eg: from `rhaos-4.0-rhel-7-candidate` to `rhaos-4.1-rhel-7-candidate` manually
+        :param pkg_name: Missing packages like: No package (.*) available
+        """
+        latest = self.metadata.exec_latest_build_info(pkg_name)
```
sosiouxme: This queries the current candidate tag for the RPM name that was missing. There are several reasons why that's never going to return anything:

1. It is the RPM name, and brew only deals in package names. They might be the same, but often they're not.
2. It might not even be a name. It could be a version lock e.g. `python-foo-3.*` or `golang >= 1.10`. It might be a requirement that's actually provided by an RPM of a completely different name.
3. More importantly, this looks for it in the current candidate tag. If it wasn't found during the build, then it's not in the current tag; if anywhere, it's in the past one.

So the missing package should be displayed whether or not you can find anything in brew that provides it, so that a human can search for it. The "try tagging this" advice would be additional, and you'd build that by querying the entire list of latest packages (`brew list-tagged --latest`) from the previous tag  (you can get MAJOR.MINOR from https://github.com/openshift/ocp-build-data/blob/openshift-4.1/group.yml#L178 and decrement MINOR to help determine the previous tag), listing the RPMs that those packages include (`brew buildinfo`), and then searching a dict of those results for the missing RPMs.

I think for the first pass you could just list everything that matched the pattern, and follow up with a PR to actually recommend what to tag to satisfy what's missing.

doozerlib/distgit.py
```diff
@@ -935,6 +937,17 @@ def _extract_container_build_logs(self, task_id):
...
+        if latest.startswith(pkg_name):
+            with self.runtime.mutex:
+                self.runtime.missing_pkgs.add(latest)
```
sosiouxme: In addition to what was missing, could the output indicate which image build wanted it? Helpful context.

doozerlib/distgit.py
shiywang: Yes, I'll remove the exec_latest_build in this pr

doozerlib/distgit.py
shiywang: Are u suggest change a name of `missing-pkgs`?

sosiouxme: The name is fine. It's just that it's going to say something like
```
Missing packages:
foo
bar
```

And it would be nice if it was more like
```
Missing packages:
openshift-enterprise-base image is missing foo
ansible-service-broker image is missing bar
```
For now each entry in missing_pkgs would be one of those lines.
```suggestion
    self.runtime.missing_pkgs.add("{} image is missing package {}".format(self.metadata.name, pkg_name))
```

shiywang:
@sosiouxme PTAL

doozer
```diff
@@ -851,6 +851,8 @@ def images_build_image(runtime, odcs, repo_type, repo, push_to_defaults, push_to
     failed = [name for name, r in results if not r]
     if failed:
         runtime.logger.error("\n".join(["Build/push failures:"] + sorted(failed)))
+        if len(runtime.missing_pkgs) != 0:
```
adammhaile: More pythonic to leave out the `!= 0`

```suggestion
        if len(runtime.missing_pkgs):
```

shiywang:
@adammhaile @sosiouxme thank you for all the reviewing,  PTAL : )

sosiouxme:
@shiywang conflict now, please rebase...

PR #70, opened 2019-03-29
Adding settings path and container_opts

adammhaile:
A couple changes required for running in production.
- Can now set `DOOZER_SETTINGS` envvar (rundoozer only) with path to yaml config to us (so I can use the jenkins settings when running as podman as root)
- For good measure, added ability to include arbitrary CLI options to the podman invocation. In case we need to mount another directory or something. Not sure if it'll be needed yet, but easy enough to add.

### DIFF
--- a/rundoozer/rundoozer
+++ b/rundoozer/rundoozer
@@ -78,6 +78,9 @@ CLI_OPTS = {
     'image': {
         'env': 'DOOZER_IMAGE',
         'help': 'Pull spec of image to use if not default'
+    },
+    'container_opts': {
+        'help': 'List of config options to send to podman when running doozer container'
     }
 }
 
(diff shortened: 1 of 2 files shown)


rundoozer/rundoozer
```diff
@@ -214,10 +217,21 @@ def main():
 
     doozer_args = clean_args(doozer_args)
 
-    cfg = dotconfig.Config('rundoozer', 'settings',
-                           template=CLI_CONFIG_TEMPLATE,
-                           envvars=CLI_ENV_VARS,
-                           cli_args=args)
+    settings_file = None
+    if 'DOOZER_SETTINGS' in os.environ:
+        settings_file = os.environ['DOOZER_SETTINGS']
```
tbielawa: Couldn't you do like....

`settings_file = os.environ.get('DOOZER_SETTINGS', None)`
?

rundoozer/rundoozer
```diff
@@ -214,10 +217,21 @@ def main():
...
+
+    try:
+        cfg = dotconfig.Config('rundoozer', 'settings',
+                               template=CLI_CONFIG_TEMPLATE,
+                               envvars=CLI_ENV_VARS,
+                               cli_args=args,
+                               file_override=settings_file)
+    except Exception as e:
+        raise
```
tbielawa: You raise, this block will exit before getting to any print part, won't it?

rundoozer/rundoozer
adammhaile: Yep... old habits. I know. I know.

rundoozer/rundoozer
adammhaile: Yep... was debug. Removed.

PR #71, opened 2019-03-29
Updating docs

adammhaile:
@tbielawa This one's for you.
I admit the docs are a little confusing now because we have both containerized and non-containerized doozer. But I've left as much of both in there as I can since they are both required. 
Will probably take another pass at this once take the plunge to only support containerized doozer.


sosiouxme:
I like docs

PR #73, opened 2019-04-02
rpmcfg.py: enable pushing release commit to origin

sosiouxme:
Ref. https://github.com/openshift/ocp-build-data/pull/74/files
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1694183

A standard doozer RPM build is less intrusive to the source repo than
the usual tito flow because the commit it uses for the build isn't
pushed back to origin; the origin commit is tagged and the tag pushed
but the release commit is discarded.

The minor problem with this is that `tito release` injects the commit
hash it releases right into the build release, and this hash will never
be present in the source repo.

It turns out that for the openshift RPM, something really cares about
that git hash being available in the source repo for validation. So this
change enables an option to push that release commit back to the source.

Also the openshift RPM has specialized tito config already. Doozer logic
already imitates it pretty well, but it seems wisest to minimize
deviations from upstream by actually using the tito config to tag and
release; so now the logic can be configured to do that.

### DIFF
--- a/doozerlib/rpmcfg.py
+++ b/doozerlib/rpmcfg.py
@@ -1,6 +1,7 @@
 import glob
 import os
 import traceback
+import re
 
 import exectools
 from pushd import Dir
@@ -11,7 +12,7 @@ from model import Missing
 from doozerlib.exceptions import DoozerFatalError
 
 RELEASERS_CONF = """
-[aos]
+[{target}]
 releaser = tito.release.DistGitReleaser
 branches = {branch}
 srpm_disttag = .el7aos
@@ -19,9 +20,10 @@ builder.test = 1
 remote_git_name = {name}
 """
 
+# note; appended to, does not replace existing props
 TITO_PROPS = """
 
-[aos]
+[{target}]
 remote_git_name = {name}
 """
 
@@ -39,7 +41,6 @@ class RPMMetadata(Metadata):
         self.version = None
         self.release = None
         self.tag = None
-        self.vtag = None
         self.commit_sha = None
         self.build_status = False
 
@@ -68,15 +69,13 @@ class RPMMetadata(Metadata):
         self.version = version
         self.release = release
         self.tag = '{}-{}-{}'.format(self.config.name, self.version, self.release)
-        self.vtag = 'v{}-{}'.format(self.version, self.release)
 
     def create_tag(self, scratch):
         if not self.tag:
             raise ValueError('Must run set_nvr() before calling!')
 
         with Dir(self.source_path):
-            exectools.cmd_assert('git tag {}'.format(self.tag))
-            exectools.cmd_assert('git tag {}'.format(self.vtag))
+            exectools.cmd_assert(["git", "tag", "-am", "Release with doozer", self.tag])
             rc, sha, err = exectools.cmd_gather('git rev-parse HEAD')
             self.commit_sha = sha.strip()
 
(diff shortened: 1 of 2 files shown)


doozerlib/rpmcfg.py
```diff
@@ -68,15 +69,13 @@ def set_nvr(self, version, release):
...
         self.tag = '{}-{}-{}'.format(self.config.name, self.version, self.release)
-        self.vtag = 'v{}-{}'.format(self.version, self.release)
 
     def create_tag(self, scratch):
         if not self.tag:
             raise ValueError('Must run set_nvr() before calling!')
 
         with Dir(self.source_path):
-            exectools.cmd_assert('git tag {}'.format(self.tag))
-            exectools.cmd_assert('git tag {}'.format(self.vtag))
```
adammhaile: why are you getting rid of vtag?

doozerlib/rpmcfg.py
```diff
@@ -85,32 +84,73 @@ def push_tag(self):
             raise ValueError('Must run set_nvr() before calling!')
 
         with Dir(self.source_path):
-            exectools.cmd_assert('git push origin {}'.format(self.tag), retries=3)
+            exectools.cmd_assert('git push origin --tags', retries=3)
 
-    def commit_changes(self):
+    def commit_changes(self, scratch):
         with Dir(self.source_path):
+            if self.config.content.build.use_source_tito_config:
```
adammhaile: curious if you should check if scratch mode here, as this would not be necessary

doozerlib/rpmcfg.py
```diff
@@ -85,32 +84,73 @@ def push_tag(self):
...
 
     def tito_setup(self):
+        if self.config.content.build.use_source_tito_config:
+            return  # rely on tito already set up in source
+
         tito_dir = os.path.join(self.source_path, '.tito')
+        tito_target = self.config.content.build.tito_target
+        tito_target = tito_target if tito_target else 'aos'
+        tito_dist = self.config.content.build.tito_dist
+        tito_dist = tito_dist if tito_dist else '.el7aos'
```
adammhaile: hard coded rhel7, should be pulled from group.yaml

doozerlib/rpmcfg.py
sosiouxme: I introduced vtag to support `openshift` RPM which has a custom tagger to do that (`openshift-ansible` may as well). Basically because I didn't understand all of doozer's interactions with tito. Since this now uses `openshift`'s tito config correctly it's no longer needed.

doozerlib/rpmcfg.py
sosiouxme: tag is still needed in a scratch build (`tito release` looks at the tag); it just doesn't push the tag and commit.

doozerlib/rpmcfg.py
```diff
@@ -85,32 +84,73 @@ def push_tag(self):
...
-    def commit_changes(self):
+    def commit_changes(self, scratch):
         with Dir(self.source_path):
+            if self.config.content.build.use_source_tito_config:
+                # just use the tito tagger to change spec and tag
+                exectools.cmd_assert(["tito", "tag", "--no-auto-changelog",
+                    "--use-version", self.version,
+                    "--use-release", "{}%{{?dist}}".format(self.release)
+                ])
+                if self.config.content.build.push_release_commit and not scratch:
```
sosiouxme: Here the commit is not pushed if it's scratch.

doozerlib/rpmcfg.py
sosiouxme: just for the default :)

I agree this ought to come from group.yaml -- all of these new attrs probably should. Just wasn't ready to introduce that extra bit of logic for the immediate fix. I guess these could be properties of `RPMMetadata`...

PR #74, opened 2019-04-02
Fix dry_run

adammhaile:
Tried to push images this morning and noticed that `--dry-run` still tried to push. Quick fix. Seriously, I only added one line, github needs to be better about showing indent on diffs :P

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -563,28 +563,31 @@ class ImageDistGitRepo(DistGitRepo):
 
                     mirror_cmd = 'oc image mirror {} {} --filename={}'.format(dr, insecure, push_config)
 
-                    for r in range(10):
-                        self.logger.info("Mirroring image [retry={}]".format(r))
-                        rc, out, err = exectools.cmd_gather(mirror_cmd)
-                        if rc == 0:
-                            break
-                        self.logger.info("Error mirroring image -- retrying in 60 seconds.\n{}".format(err))
-                        time.sleep(60)
-
-                    lstate = self.runtime.state[self.runtime.command] if self.runtime.command == 'images:push' else None
-
-                    if rc != 0:
-                        if lstate:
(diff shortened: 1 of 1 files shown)


doozerlib/distgit.py
```diff
@@ -563,28 +563,29 @@ def push_image(self, tag_list, push_to_defaults, additional_registries=[], versi
...
-                    if rc != 0:
-                        if lstate:
-                            state.record_image_fail(lstate, self.metadata, 'Build failure', self.runtime.logger)
-                        # Unable to push to registry
-                        raise IOError('Error pushing image: {}'.format(err))
-                    else:
-                        if lstate:
-                            state.record_image_success(lstate, self.metadata)
-                        self.logger.info('Success mirroring image')
+                    if not dry_run: # skip everything else if dry run
```
sosiouxme: That's what `Hide whitespace changes` is for under diff settings...

Although it would be more pythonic here to:
```suggestion
                    if dry_run:
                        continue  # skip everything else for this image
```

adammhaile:
@sosiouxme fixed your suggestion

doozerlib/distgit.py
```diff
@@ -563,28 +563,31 @@ def push_image(self, tag_list, push_to_defaults, additional_registries=[], versi
...
-                    lstate = self.runtime.state[self.runtime.command] if self.runtime.command == 'images:push' else None
-
-                    if rc != 0:
-                        if lstate:
-                            state.record_image_fail(lstate, self.metadata, 'Build failure', self.runtime.logger)
-                        # Unable to push to registry
-                        raise IOError('Error pushing image: {}'.format(err))
+                    if dry_run:  # skip everything else if dry run
+                        continue
                     else:
```
sosiouxme: you don't need the else though, because the continue skips the rest... was trying to save you the massive whitespace change :)

PR #77, opened 2019-04-09
rpmcfg: use tito tag or doozer tag, not both

sosiouxme:
The code was confusing enough that I didn't notice in the case of `use_source_tito_config` it was tagging it twice. It didn't break on `openshift` because that uses a custom tagger and the tag doesn't conflict.

Along the way I noticed some cruft and bits that should be much clearer now.

### DIFF
--- a/doozerlib/rpmcfg.py
+++ b/doozerlib/rpmcfg.py
@@ -41,7 +41,6 @@ class RPMMetadata(Metadata):
         self.version = None
         self.release = None
         self.tag = None
-        self.commit_sha = None
         self.build_status = False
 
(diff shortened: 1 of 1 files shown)


sosiouxme:
I think you have a point @shiywang ... it's a pythonism to try to avoid indenting huge chunks of code, so often a "short-circuit" looks something like:
```
if foo:
    # ... do something ...
    return / break / continue
# ... continue with not-foo case ...
```
In this case, I think the first case goes on long enough that it would probably make sense to break these out into methods or something. But I'd like to consider that for a follow-up PR.

PR #78, opened 2019-04-10
add config:gen-csv new subcommand to dump csv file

shiywang:
adding an option `--write-csv` to dump csv file into `./` I can add an `--output-file` option to specify file location as well, be aware this only works with `--key` and at this time only for sort out missing image owners files.

the docs could be like this:
excel

### DIFF
--- a/doozer
+++ b/doozer
@@ -1454,6 +1454,39 @@ def config_print(runtime, key, name_only, as_yaml):
     config.config_print(key, name_only, as_yaml)
 
 
+@cli.command("config:gen-csv", short_help="Generate .csv file for given images/rpms")
+@click.option("--keys", help="Specific key in config to print, separated by commas: --keys key,name,owners",
+              default=None)
+@click.option("--type", "as_type", default=None, help='Write content type: image or rpm')
+@click.option("--output", "-o", default=None, help='Write csv data to FILE instead of STDOUT')
+@pass_runtime
+def config_gencsv(runtime, keys, as_type, output):
+    """Generate .csv file for given --keys and --type
+
+    By default print out with STDOUT, you can use --output/-o to specify an output file
+
+    Filtering of configs is based on usage of the following global options:
+    --group, --images/-i, --rpms/-r
+
+    See `doozer --help` for more.
+
+    Examples:
+
+    Generate a CSV where each row is distgit key, image name, and owners list:
+
+        $ doozer --group=openshift-4.0 config:gen-csv --type image --keys key,name,owners
+
+    Generate a CSV only include image aos3-installation each row is distgit key, image name, output to file ./image.csv:
+
+        $ doozer --group=openshift-4.0 -i aos3-installation config:gen-csv --type image --keys key,name --output ./image.csv
+
+    """
+    _fix_runtime_mode(runtime)
+    runtime.initialize(**CONFIG_RUNTIME_OPTS)
+    config = mdc(runtime)
+    config.config_gen_csv(keys, as_type, output)
+
+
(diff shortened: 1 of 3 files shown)


adammhaile:
I like where you are going with it, but see email I just sent... we can make this more generic.

shiywang:
@adammhaile PTAL

doozer
```diff
@@ -1454,6 +1454,36 @@ def config_print(runtime, key, name_only, as_yaml):
     config.config_print(key, name_only, as_yaml)
 
 
+@cli.command("config:gen-csv", short_help="Generate .csv file for given images/rpms")
+@click.option("--fields", help="Specific key in config to print, separated by commas: --fields key,name,owners",
```
adammhaile: Now that I'm looking at this I feel like `--keys` would be more in line with other usages in doozer.

doozer
```diff
@@ -1454,6 +1454,36 @@ def config_print(runtime, key, name_only, as_yaml):
...
+              default=None)
+@click.option("--type", "as_type", default=None, help='Write content type: image or rpm')
```
adammhaile: Need to add `--output / -o` option to specify output file name and directory. Without reading the code it's not obvious where this outputs to. If `--output` is not specified, just emit to `stdout`.

doozerlib/config.py
```diff
@@ -114,6 +115,39 @@ def _do_print(kind, output):
...
+            image_list = self.runtime.rpm_metas()
+        else:
+            print('Not correct --type specified (--type image or --type rpm). '
+                  'Or not consist with global options: --images/-i and --rpms/-r')
+
+        # split --fields=a,b,c,d to list [a,b,c,d]
+        fields = [c.strip() for c in fields.split(',')]
+
+        if image_list:
+            with open('{}.csv'.format(as_type), mode='w') as csv_file:
```
adammhaile: as noted above, take a user given output path or just emit to `stdout`

shiywang:
@adammhaile updated, PTAL

doozerlib/config.py
```diff
@@ -114,6 +115,51 @@ def _do_print(kind, output):
...
+            image_list = self.runtime.image_metas()
+        if as_type == "rpm" and not self.runtime.images:
+            image_list = self.runtime.rpm_metas()
+
+        if image_list is None:
+            print('Not correct --type specified (--type image or --type rpm). '
+                  'Or not consist with global options: --images/-i and --rpms/-r')
+            return
+
+        def _write_row(w):
```
adammhaile: Semantic issue... should be `_write_rows(w)` as it writes multiple rows, not just one.

shiywang: done

doozerlib/config.py
```diff
@@ -114,6 +115,51 @@ def _do_print(kind, output):
...
+
+        image_list = None
+        if as_type == "image" and not self.runtime.rpms:
+            image_list = self.runtime.image_metas()
+        if as_type == "rpm" and not self.runtime.images:
+            image_list = self.runtime.rpm_metas()
+
+        if image_list is None:
+            print('Not correct --type specified (--type image or --type rpm). '
+                  'Or not consist with global options: --images/-i and --rpms/-r')
```
sosiouxme: Tiny nitpick:
```suggestion
                  'Or not consistent with global options: --images/-i and --rpms/-r')
```
Otherwise, LGTM!

shiywang:
@sosiouxme PTAL

sosiouxme:
Sorry to nitpick you for several days :) And thanks Adam for doing most of the reviewing so I can just say
/lgtm

PR #79, opened 2019-04-12
distgit.py: search builds for component not dg name

sosiouxme:
When doing a scan for changed sources, the test for whether the distgit
copy of the source has been built looks for brew builds. But it was
looking for the distgit name instead of the component. For RPMs they are
the same but images generally add -container or -apb or explicitly
specify a component.


sosiouxme:
There's no urgent need to update buildvm for this; incrementals just kind of become daily full builds without it. We can live with that until monday.

PR #83, opened 2019-04-26
distgit.py: work around OSBS bug

sosiouxme:
This is an ugly hack to work around an OSBS bug regarding multiline statements.
For now, combine into single line.
Revert this once a fix appears.

I did test this on a few rebases to see that it worked...


tbielawa:
/lgtm
seems legit. Let's give it a shot and see if it works!

PR #91, opened 2019-05-09
fixing os.mkdir race-condition in push_image

shiywang:
the problem I assume happened in here: https://buildvm.openshift.eng.bos.redhat.com:8443/job/aos-cd-builds/job/build%252Fbuildvm-maint/618/consoleFull

Not sure if it's a pythonic way, @adammhaile PTAL


shiywang:
@adammhaile sure will do that.

shiywang:
@adammhaile I merged the pr, but we have to make a release to actually sync up to buildvm right? (I am not in a hurry just a question)

adammhaile:
Yep. I'll do that later today.

PR #92, opened 2019-05-09
Fix doozer build lock failure

adammhaile:


sosiouxme:
Doesn't this just put it back to the way it was before https://github.com/openshift/doozer/pull/87/files ?

adammhaile:
@sosiouxme yep, that's exactly what it does. The problem was actually elsewhere (it was building images twice) but the issue manifested as looking like it was part of the build lock. Once I fixed the real problem the original code was actually valid.

PR #94, opened 2019-05-13
distgit: add envvars to relay source metadata

sosiouxme:
urgent last-minute fix for MCO and https://jira.coreos.com/browse/ART-460


sosiouxme:
Solely waiting to see if clayton wants different var name(s)

PR #111, opened 2019-06-05
Cover image distgit push image

thiagoalessio:
Captured all possible variations/outcomes of `ImageDistGitRepo`'s `push_image` method in unit tests.

#### before
```
py27 run-test: commands[0] | flake8
py27 run-test: commands[1] | coverage run --branch --source doozerlib -m unittest discover
s...No handlers could be found for logger "doozer.tests.test_runtime"
..s..................s.s.........................................s.s.s.s....s.s.
----------------------------------------------------------------------
Ran 74 tests in 0.282s

OK (skipped=10)
py27 run-test: commands[2] | coverage report
Name                      Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------
...
doozerlib/distgit.py        989    674    458      3    29%
...
-----------------------------------------------------------
TOTAL                      2691   1926   1105     14    24%
```

#### after
```
py27 run-test: commands[0] | flake8
py27 run-test: commands[1] | coverage run --branch --source doozerlib -m unittest discover
s...No handlers could be found for logger "doozer.tests.test_runtime"
..s...........................................................s.s.............s.s.s.s....s.s.
----------------------------------------------------------------------
Ran 87 tests in 0.332s

OK (skipped=10)
py27 run-test: commands[2] | coverage report
Name                      Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------
...
doozerlib/distgit.py        989    593    458      3    37%
...
-----------------------------------------------------------
TOTAL                      2691   1845   1105     14    27%
```

PR #112, opened 2019-06-06
Use only flexmock in test_generic_distgit

thiagoalessio:
As discussed in our arch meeting, i'm getting rid of all usages of `mock` in favor of `flexmock`.
~~This PR only covers `test_generic_distgit.py`.~~ This PR covers everything I had done previously **except** the contents of #111, because they are still not on master. But i'm already working on it ;D

### DIFF
--- a/tests/test_distgit/test_generic_distgit.py
+++ b/tests/test_distgit/test_generic_distgit.py
@@ -1,12 +1,14 @@
 import errno
+import os
 import unittest
-import mock
 from datetime import datetime, timedelta
+
+import flexmock
 
 import distgit
 from model import Model
 
-from .support import MockMetadata, MockRuntime, TestDistgit
+from .support import MockConfig, MockMetadata, MockRuntime, TestDistgit
 
 
 class TestGenericDistGit(TestDistgit):
@@ -22,10 +24,11 @@ class TestGenericDistGit(TestDistgit):
         self.assertIsInstance(self.dg, distgit.DistGitRepo)
 
     def test_init_with_branch_override(self):
-        metadata = mock.Mock()
-        metadata.runtime.branch = "original-branch"
+        metadata = flexmock(runtime=flexmock(branch="original-branch"),
+                            config=flexmock(distgit=flexmock(branch=distgit.Missing)),
+                            name="_irrelevant_",
+                            logger="_irrelevant_")
 
-        metadata.config.distgit.branch = distgit.Missing
         repo = distgit.DistGitRepo(metadata, autoclone=False)
         self.assertEqual("original-branch", repo.branch)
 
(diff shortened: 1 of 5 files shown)


tests/test_distgit/test_generic_distgit.py
```diff
@@ -22,194 +24,286 @@ def test_init(self):
         self.assertIsInstance(self.dg, distgit.DistGitRepo)
 
     def test_init_with_branch_override(self):
-        metadata = mock.Mock()
-        metadata.runtime.branch = "original-branch"
+        metadata = flexmock(runtime=flexmock(branch="original-branch"),
+                            config=flexmock(distgit=flexmock(branch=distgit.Missing)),
```
thiagoalessio: @sosiouxme and/or @adammhaile I'd like to ask if you are aware of a more elegant way of defining nested **properties** with flexmock, because this looks UGLY! :nauseated_face: 

I've seen that they offer a nicer API for nested **method** calls, something like:
```python
flexmock(x).should_receive('foo.bar.baz')
# x.foo().bar().baz()
```
but i couldn't find an equivalent for **properties** :disappointed: any advice ?

tests/test_distgit/test_generic_distgit.py
```diff
@@ -22,194 +24,286 @@ def test_init(self):
...
+                            name="_irrelevant_",
+                            logger="_irrelevant_")
```
thiagoalessio: That's something else I dislike on my implementation ... I see the value of not hiding anything, and it resonates with the Zen of Python:
> Explicit is better than implicit.

but I feel it pollutes the code and hides the properties that are actually relevant for the current test.
What are your thoughts in this matter ?

tests/test_distgit/test_generic_distgit.py
adammhaile: Not that I'm aware of

tests/test_distgit/test_generic_distgit.py
adammhaile: I'm all for clean but understandable. Personally I feel some python programmers go overboard with explicit over implicit... I'd rather it be readable.

adammhaile:
Looks like you are still working on this anyways, but same comment as other PR regarding squashing before merge.

thiagoalessio:
@adammhaile i'm done actually ... i just thought that i could add some more commits this morning before you woke up ;D

adammhaile:
@thiagoalessio In that case - /lgtm but please squash before merging

tests/test_distgit/test_generic_distgit.py
thiagoalessio: great, i also share this opinion. i'll try to come up with something cleaner soon.

PR #115, opened 2019-10-11
Move CI code to shared library

thiagoalessio:
Loaded from https://github.com/thiagoalessio/art-ci-toolkit (temporarily, if it works we should request a new openshift org repo)


thiagoalessio:
PR was just for testing, actual code is already living in master ;D

PR #117, opened 2019-10-11
ART-948: Support specifying out-of-tree sources in ocp-build-data.

sosiouxme:
copied over MR 145 from gitlab to have automation test it


vfreex:
Fixed commit were sent to Gitlab.
BTW it is interesting that different versions of flake8 report different results.

PR #119, opened 2019-10-16
Fix operator metadata default channel

thiagoalessio:


sosiouxme:
LGTM

thiagoalessio:
crazy diff coverage

PR #121, opened 2019-10-17
Allow group.yml to disable builds

jupierce:
`freeze_automation: yes`  - prevent all builds, including from command line (enforced by doozer)
`freeze_automation: no` - don't prevent any builds 
`freeze_automation: scheduled` - prevent scheduled builds, but nothing manual (enforced by pipeline code which should query doozer group with config:read-group)

### DIFF
--- a/doozer
+++ b/doozer
@@ -232,6 +232,7 @@ def rpms_build(runtime, version, release, scratch):
         version = version[1:]
 
     runtime.initialize(mode='rpms', clone_distgits=False)
+    runtime.assert_mutation_is_permitted()
 
     items = runtime.rpm_metas()
     if not items:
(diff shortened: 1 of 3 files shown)


sosiouxme:
1. `image:rebase` should also check whether builds are permitted
2. the group.yml schema isn't really used by anything AFAIK -- our validation is in ocp-build-data-validator so that at least would need an update here.

jupierce:
@sosiouxme I don't see any group validation in https://github.com/openshift/ocp-build-data-validator .

vfreex:
@jupierce I think this PR belongs to what we talked at the Arch meeting?
Regarding RFEs, will you open other PRs or update this PR?

sosiouxme:
Alright, I think it's good. We'll find out soon enough if there's any actual validation on group.yml schema...

PR #123, opened 2019-10-21
[ART-1005] operator-metadata: work with NVR

joepvd:
With the advent of ART-874 the process of actually shipping these new metadata containers is fairly cumbersome and error-prone. With every release slightly overlapping with the ones before and after, it's easy to include metadata for the wrong operator containers.

This PR is for improvements to make it easier to include the right metadata and verify it afterward:

- `doozer operator-metadata:build` takes an NVR and a stream; currently it creates a container with the same V and auto-incrementing R starting at 1. With several operator builds at the same version it's tedious to determine which is intended. Let's make them easily identifiable. The metadata container still needs the R to auto-increment, but we can put whatever we want in the V, so let's put the original R and the stream in there:
  `elasticsearch-operator-container-v4.1.16-201909131234`
  gets metadata container:
  `elasticsearch-operator-metadata-container-v4.1.16.201909131234.dev-1`
- `doozer operator-metadata:latest-build` takes a distgit and looks up the latest metadata container on the current openshift version. It needs to be able to look up the latest metadata container for a specific operator NVR and stream.
  `doozer operator-metadata:latest-build --nvr elasticsearch-operator-container-v4.1.16-201909131234 --stream stage`
  gets metadata container:
  `elasticsearch-operator-metadata-container-v4.1.16.201909131234.stage-<1,2,3...>`

### DIFF
--- a/doozer
+++ b/doozer
@@ -2187,9 +2187,9 @@ def update_operator_metadata(runtime, nvr_list, stream, merge_branch, force=Fals
 
 
 @cli.command("operator-metadata:latest-build", short_help="Print latest metadata build of a given operator")
-@click.option("--nvr", "-n", help="Specify Name-Version-Release", default=None, required=False)
+@click.option("--nvr", "-n", help="Specify Name-Version-Release. Needs --stream", multiple=True)
 @click.option("--stream", "-s", type=click.Choice(['dev', 'stage', 'prod']),
-    help="Metadata stream to select. Possible values: dev, stage, prod.")
+    help="Metadata stream to select. Possible values: dev, stage, prod. Needs --nvr")
 @click.argument("operator_list", nargs=-1)
 @pass_runtime
 def operator_metadata_latest_build(runtime, nvr, stream, operator_list):
@@ -2207,7 +2207,7 @@ def operator_metadata_latest_build(runtime, nvr, stream, operator_list):
 
     Usage example when feeding NVR and stream:
 
-    $ doozer -g openshift-4.2 operator-metadata:latest-build --nvr cluster-logging-operator-container-v4.2.1-201910221723 --stream stage
+    $ doozer -g openshift-4.2 operator-metadata:latest-build --stream stage --nvr cluster-logging-operator-container-v4.2.1-201910221723 --nvr sriov-network-operator-container-v4.2.0-201909131819
     """
 
     if operator_list and (stream or nvr):
@@ -2219,14 +2219,18 @@ def operator_metadata_latest_build(runtime, nvr, stream, operator_list):
     if nvr and not stream:
         click.echo('--nvr needs to be accompanied by --stream. Exiting.', err=True)
         sys.exit(1)
+    if not nvr and not operator_list:
+        click.echo("Missing arguments: Need nvr's or operators to work with")
+        sys.exit(1)
 
     runtime.initialize(clone_distgits=False)
 
     if operator_list:
         for operator in operator_list:
-            print(operator_metadata.OperatorMetadataLatestBuildReporter(operator, runtime).get_latest_build())
+            click.echo(operator_metadata.OperatorMetadataLatestBuildReporter(operator, runtime).get_latest_build())
     else:
-        print(operator_metadata.OperatorMetadataLatestNvrReporter(nvr, stream, runtime).get_latest_build())
+        for build in nvr:
+            click.echo(operator_metadata.OperatorMetadataLatestNvrReporter(build, stream, runtime).get_latest_build())
 
 
 def main():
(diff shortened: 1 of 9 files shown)


doozer
```diff
@@ -2187,9 +2187,12 @@ def update_operator_metadata(runtime, nvr_list, stream, merge_branch, force=Fals
 
 
 @cli.command("operator-metadata:latest-build", short_help="Print latest metadata build of a given operator")
-@click.argument("operator_list", nargs=-1, required=True)
+@click.option("--nvr", "-n", help="Specify Name-Version-Release", default=None, required=False)
```
thiagoalessio: @sosiouxme was that your intention? When I read the ticket I also understand that it should receive a single NVR, but wouldn't be better to receive multiple ones? (maybe a RFE for the future)

sosiouxme: it really should be able to specify `--nvr` multiple times and look up all at once rather than have to invoke it for each, if for no other reason than for consistency, but it's not critical right now unless you need it for the jenkins work...

doozerlib/operator_metadata.py
```diff
@@ -260,10 +258,12 @@ def create_metadata_dockerfile(self):
         )
         metadata_dockerfile.labels['com.redhat.delivery.appregistry'] = 'true'
         metadata_dockerfile.labels['name'] = 'openshift/ose-{}'.format(self.metadata_name)
-        try:
-            del(metadata_dockerfile.labels['release'])
-        except KeyError:
-            pass
+        # mangle version according to spec
+        metadata_dockerfile.labels['version'] = '{}-{}.{}'.format(
```
sosiouxme: ... have you tried running a build with this? I didn't think you could have a `-` in that label. I expected it to be: 
```suggestion
        metadata_dockerfile.labels['version'] = '{}.{}.{}'.format(
```

doozerlib/operator_metadata.py
```diff
@@ -575,6 +575,65 @@ def operator(self):
...
+            # the new naming scheme. https://jira.coreos.com/browse/ART-1175
+            logger.warning('Did not find a match under the new naming scheme. Falling back to the old')
+            candidate = OperatorMetadataLatestBuildReporter(self.operator_name, self.runtime).get_latest_build()
+
+        return candidate
+
+    @log
+    def get_all_builds(self):
+        """Ask brew for all releases of a package"""
+        cmd = 'brew list-builds --package {} --state COMPLETE --quiet'.format(self.metadata_name)
```
sosiouxme: hmm... this gives you all the package builds ever for any release...
`brew list-tagged rhaos-4.2-rhel-7-candidate  cluster-logging-operator-metadata-container` would give you just the ones that are eligible for 4.2
I could see some situations where what you have might bring up something that's not actually tagged and can't be added to the advisory.

doozerlib/operator_metadata.py
```diff
@@ -575,6 +575,65 @@ def operator(self):
...
+    def __init__(self, operator_nvr, stream, runtime):
+        self.operator_nvr = operator_nvr
+        self.stream = stream
+        self.operator_name, self.operator_version = self.unpack_operator_nvr()
+        self.metadata_name = '{}-metadata-container'.format(self.operator_name)
+        self.runtime = runtime
+
+    @log
+    def get_latest_build(self):
+        regex = r'^{}-metadata-container-{}\.{}-(?P<release>[0-9]+)$'.format(
```
sosiouxme: Given above I would think this needs to change too

doozerlib/operator_metadata.py
```diff
@@ -575,6 +575,65 @@ def operator(self):
...
+
+        for line in stdout.splitlines():
+            yield line.split(' ')[0]
+
+    def ask_brew(self, cmd):
+        _rc, stdout, _stderr = exectools.cmd_gather(cmd)
+        return stdout
+
+    def unpack_operator_nvr(self):
+        unpacked = re.match(r'(?P<operator>.*)-container-(?P<version>v[0-9]+\.[0-9]+\.[0-9]+-[0-9]+)', self.operator_nvr)
```
sosiouxme: i really don't like to attempt an NVR regex. For instance this assumes a specific version format. If you just `split('-')` then the last two are the VR and the rest combined are the N.

joepvd: Thanks!

doozerlib/operator_metadata.py
joepvd: That's a better approach. Thanks!

doozerlib/operator_metadata.py
thiagoalessio: solved here: https://github.com/openshift/doozer/pull/123/commits/066a47889b068316af499db4b49e5c9b869acb7a#diff-9206bb1bd248a870ebf5892a4dab3341R595-R602

PR #129, opened 2019-10-30
Allow doozer to filter multiple spec files

jupierce:


dulek:
Makes sense to me!

sosiouxme:
LGTM

PR #134, opened 2019-11-03
Embed pygitdata for future refactoring and Python 3 migration

vfreex:


thiagoalessio:
/lgtm

just out of curiosity, do you plan to also embed "dotconfig" ?
https://github.com/adammhaile/dotconfig

vfreex:
@thiagoalessio It has already happened: https://github.com/openshift/doozer/pull/127
It is good if you didn't notice that (which means I didn't blow things up) 😏

PR #140, opened 2019-11-09
Python 3 porting: Run `futurize --stage1`

vfreex:
I am trying [Futurize][] for porting Python 2 code to Python 3.
The tool is one of recommended tools listed on the official Python HOWTOs documentation [Porting Python 2 Code to Python 3][].

Futurize can divide the porting into two stages. Stage 1 is for “safe” changes that modernize the code but do not break Python 2.6 compatibility or introduce a depdendency on the future package. Stage 2 is to complete the process.

The PR is generated with the following command:

```bash
futurize --stage1 --unicode-literals --nobackups --write .
```

[Porting Python 2 Code to Python 3]: https://docs.python.org/3/howto/pyporting.html
[Futurize]: http://python-future.org/automatic_conversion.html

## NOTE
1. Only stage 1 of the Futurize porting process is run so that we are expecting this will not break our existing Python 2 compatibility but it will still not run on Python 3.

2. Python 3 doesn't allow to unpack tuple argument in a lambdas or functions (PEP-3113). e.g.

```python
results = runtime.parallel_exec(
        lambda (rpm, terminate_event): rpm.build_rpm(
            version, release, terminate_event, scratch),
        items)
```

This could be change to use multiple arguments:

``` python
results = runtime.parallel_exec(
        lambda rpm, terminate_event: rpm.build_rpm(
            version, release, terminate_event, scratch),
        items)
```

However, `ThreadPool.map_async` only accept functions with a single argument.

To workaround this, I added a wrapper function `_unpack_tuple_args` to unpack tuple as arguments for the function passed to `ThreadPool.map_async`.

NOTE: In the future when we don't keep compatibility with Python 2, `starmap_async` can be used instead.

### DIFF
--- a/doozerlib/__init__.py
+++ b/doozerlib/__init__.py
@@ -1,3 +1,4 @@
+from __future__ import unicode_literals
 from .runtime import Runtime
 from .pushd import Dir
 
--- a/doozerlib/assertion.py
+++ b/doozerlib/assertion.py
@@ -5,6 +5,7 @@ the asserted condition is not met.
 The use of the FileNotFound exception makes this Python3 ready.
 Making them functions keeps the exception definition localized.
 """
+from __future__ import unicode_literals
 import os
 import errno
 
--- a/doozerlib/brew.py
+++ b/doozerlib/brew.py
@@ -1,6 +1,8 @@
 """
 Utility functions for general interactions with Brew and Builds
 """
+from __future__ import absolute_import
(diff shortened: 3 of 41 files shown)


sosiouxme:
Similar comments as https://github.com/openshift/elliott/pull/82

vfreex:
I made a comment there.

PR #141, opened 2019-11-11
Fix doozerlib.cli is not included in setup.py

vfreex:
In setup.py, call `find_packages` to auto-discover packages and use `exclude` parameter to exclude unit tests from installation.


thiagoalessio:
just waiting CI to finish then i'll merge it and bump the version ;D

vfreex:
Removed `__init__.py` from `tests` so that find_packages will not include unit tests for installation.

thiagoalessio:
oooh @vfreex removing `__init__.py` from tests prevents tox to find them:
```
py27 run-test: commands[1] | coverage run --branch --source doozer,doozerlib -m unittest discover

----------------------------------------------------------------------
Ran 0 tests in 0.000s
```

setup.py
```diff
@@ -20,7 +20,7 @@ def _get_version():
     long_description=open('README.md').read(),
     url="https://github.com/openshift/doozer",
     license="Apache License, Version 2.0",
-    packages=["doozerlib"],
+    packages=find_packages(),
```
thiagoalessio:
```suggestion
    packages=find_packages(exclude=("tests",)),
```

thiagoalessio: so you don't need to get rid of `__init__.py` on test modules.

vfreex: Thanks for the hint. I did some tests and found `test.*` should also be specified in order to exclude subpackages.

PR #142, opened 2019-11-12
[ART-1245] runtime: leave brew_tag None by default

sosiouxme:
Per https://jira.coreos.com/browse/ART-1245
This allows us to continue overriding the tag by specifying the branch in configs.

### DIFF
--- a/doozerlib/runtime.py
+++ b/doozerlib/runtime.py
@@ -342,9 +342,8 @@ class Runtime(object):
                 scanner.matches = regexen
 
             if self.group_config.brew_tag:
+                # setting this overrides tags made out of branches in specific configs
                 self.brew_tag = self.group_config.brew_tag
-            else:
-                self.brew_tag = '{}-candidate'.format(self.branch)
 
             # Flattens a list like like [ 'x', 'y,z' ] into [ 'x.yml', 'y.yml', 'z.yml' ]
             # for later checking we need to remove from the lists, but they are tuples. Clone to list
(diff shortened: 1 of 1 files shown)


adammhaile:
@sosiouxme I don't have access to that Jira anymore can you please give me some context? You removed that one else statement which seems critical to the change I made... was there a bug?

sosiouxme:
@adammhaile yes there was a bug, and it's pretty serious.
Beginning with 4.2 we have several images (and now pkgs) that are building against rhel8 instead of (what is configured for the group) rhel7. For example:
https://github.com/openshift/ocp-build-data/blob/52ddd84f0adbdd8bca7b1e82197ec6e00ffa7a82/images/ironic-static-ip-manager.yml#L10

The problem is that with your change, this gets ignored when looking for a build in brew. Only the tag specified in the group or on the command line is used. So for that image, once it got built, doozer would fail to find it because it's looking in the `rhaos-4.3-rhel-7-candidate` tag instead of `rhaos-4.3-rhel-8-candidate` where it is.

This change makes it so that doozer still uses the tag specified via group or cmdline, but leaves it `None` if not specifically overridden on the cmdline because that way the branch in the config can override it (or it falls back to what's in the group).

adammhaile:
As far as I knew the tag was never None before and my update made it behave
exactly as before unless brew_tag was set in yaml or the CLI. That was
incorrect? How would it have worked if set to None? It needs to validate
the image is in brew after build and it needs a tag to check against.

sosiouxme:
Note that my changes are in the runtime. There never was a `brew_tag` there before; you added it with https://github.com/openshift/doozer/commit/7628c16835743dd010dd61b178a0305c52402be1#diff-243066ce9bb9c9bd0f48a8e508bab1f5R122
and immediately ensured it would always be defined with https://github.com/openshift/doozer/commit/7628c16835743dd010dd61b178a0305c52402be1#diff-243066ce9bb9c9bd0f48a8e508bab1f5R344-R347

This breaks the logic you added in metadata https://github.com/openshift/doozer/commit/7628c16835743dd010dd61b178a0305c52402be1#diff-a85d43c8cb55d2cade49c1defcea9394R108-R112
The `else` there can never be reached because there's always a `self.runtime.brew_tag`. So I just made that not be the case, unless you specifically set a tag on the runtime.

adammhaile:
Gotcha... ok, it was as clear with just the diff. But that makes sense now.
/lgtm

PR #143, opened 2019-11-14
Allow images missing for different arches

jupierce:
Goes hand in hand with https://github.com/openshift/aos-cd-jobs/pull/2035


tbielawa:
I'll get testing this today.

tbielawa:
This is real close! I just noticed a few little differences from the existing output and a note I'd like added:

* Could you add to the command docstring `If you use -i,--images then you must manually include 'openshift-enterprise-cli' in the list`
* The pushed tags do not follow the same convention as we have been following up until now. Previously we pushed tags like this

> quay.io/openshift-release-dev/ocp-v4.0-art-dev:v4.3.0-201911132228-s390x-ose-cli
> quay.io/openshift-release-dev/ocp-v4.0-art-dev:v4.3.0-201911132228-ose-cli
VERSION-RELEASE-[ARCH-]ose-COMPONENT

This PR has changed that. The `ose` prefix has been omitted:
> quay.io/openshift-release-dev/ocp-v4.0-art-dev:v4.3.0-201911132228-s390x-cli

This appears in the SRC=DEST files as well as the Image Streams (which is consistent). That format has been defined for a while, I can't think of any reason we should make that change right now.

I noticed the SRC=DEST files are referencing SRC images by SHA digest now:

> registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-machine-config-operator@sha256:a16301e0b459bb2dda1d22a21b776536804147008544a4738b603397f8601f31=quay.io/openshift-release-dev/ocp-v4.0-art-dev:v4.3.0-201911140352-s390x-machine-config-operator

Which isn't necessarily a problem, it's just different and caught my eye. To satisfy my own curiosity, is there a particular reason for this change, or did it just work out that way when you wrote the patch?

jupierce:
- Added ose- back in. 
- By using sha, we eliminate the dependency on `oc mirror image` to filter by arch. The shas in the mirroring list are precisely the right image for a specific arch (not manifest lists). In fact, all the src_dest files could be in a single file now if we wanted.

tbielawa:
> * Added ose- back in.

:+1: 

> * By using sha, we eliminate the dependency on `oc mirror image` to filter by arch. The shas in the mirroring list are precisely the right image for a specific arch (not manifest lists).

Nice. Super smart.

> In fact, all the src_dest files could be in a single file now if we wanted.

:thinking: Interesting. Let's skip that for now. I think once you get the lint fixed up we're good to merge this.

PR #146, opened 2019-11-20
Fix operator metadata lookups [ART-1238] [ART-1175]

sosiouxme:
Turns out the package name was being truncated so the fallback method of lookup was always used (just get the latest in the stream).

```
$ doozer -g openshift-4.3 operator-metadata:latest-build --nvr cluster-logging-operator-container-v4.3.0-201911151317 --stream dev
...
cluster-logging-operator-metadata-container-v4.3.0.201911191807.dev-2
```

Now it correctly finds `cluster-logging-operator-container-v4.3.0-201911151317` as well as handling ptp-operator and its oddball naming correctly.

I wonder if this affected any stage pushes? It seems unlikely, but it would come into play if we ever needed to revert opmeta to refer to something earlier.

Also could not help fixing a few fiddly things along the way...

### DIFF
--- a/doozerlib/operator_metadata.py
+++ b/doozerlib/operator_metadata.py
@@ -584,13 +584,12 @@ class OperatorMetadataLatestNvrReporter:
     def __init__(self, operator_nvr, stream, runtime):
         self.operator_nvr = operator_nvr
         self.stream = stream
-
-        self.operator_name, self.operator_version, self.operator_release = self.unpack_nvr(operator_nvr)
-
-        self.metadata_name = '{}-metadata'.format(self.operator_name)
-        self.metadata_version = '{}.{}.{}'.format(self.operator_version, self.operator_release, self.stream)
-
         self.runtime = runtime
+
+        self.operator_component, self.operator_version, self.operator_release = self.unpack_nvr(operator_nvr)
+
+        self.metadata_component = self.operator_component.replace('operator-container', 'operator-metadata-container')
+        self.metadata_version = '{}.{}.{}'.format(self.operator_version, self.operator_release, self.stream)
 
     @log
     def get_latest_build(self):
@@ -598,8 +597,8 @@ class OperatorMetadataLatestNvrReporter:
         candidate = None
 
         for brew_build in self.get_all_builds():
-            name, version, release = self.unpack_nvr(brew_build)
-            if name == self.metadata_name and version == self.metadata_version and release > candidate_release:
+            component, version, release = self.unpack_nvr(brew_build)
+            if component == self.metadata_component and version == self.metadata_version and release > candidate_release:
                 candidate_release = release
                 candidate = brew_build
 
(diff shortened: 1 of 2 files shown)


sosiouxme:
(test is wrong, will fix later)

doozerlib/operator_metadata.py
```diff
@@ -584,46 +584,40 @@ class OperatorMetadataLatestNvrReporter:
     def __init__(self, operator_nvr, stream, runtime):
         self.operator_nvr = operator_nvr
         self.stream = stream
+        self.runtime = runtime
 
-        self.operator_name, self.operator_version, self.operator_release = self.unpack_nvr(operator_nvr)
+        self.operator_component, self.operator_version, self.operator_release = self.unpack_nvr(operator_nvr)
 
-        self.metadata_name = '{}-metadata'.format(self.operator_name)
+        self.metadata_component = self.operator_component.replace('operator-container', 'operator-metadata-container')
```
vfreex: Replacing `-container` with `-metadata-container` seems to be more robust without considering whether the operator name is ending with `-operator`  (is that possible?)

doozerlib/operator_metadata.py
```diff
@@ -584,46 +584,40 @@ class OperatorMetadataLatestNvrReporter:
...
-        candidate_release = 0
+        candidate_release = -1
         candidate = None
 
         for brew_build in self.get_all_builds():
-            name, version, release = self.unpack_nvr(brew_build)
-            if name == self.metadata_name and version == self.metadata_version and release > candidate_release:
+            component, version, release = self.unpack_nvr(brew_build)
+            release = int(re.search(r'\d+', release).group())
+            if component == self.metadata_component and version == self.metadata_version and release > candidate_release:
```
vfreex: Comparing only the number portion of the release value doesn't seem to be correct. Following output is an example list of metadata builds:

```
cluster-logging-operator-metadata-container-v4.2.8.201911190952.dev-1  rhaos-4.2-rhel-7-candidate  ocp-build/buildvm.openshift.eng.bos.redhat.com
cluster-logging-operator-metadata-container-v4.2.8.201911190952.dev-2  rhaos-4.2-rhel-7-candidate  ocp-build/buildvm.openshift.eng.bos.redhat.com
cluster-logging-operator-metadata-container-v4.2.8.201911190952.stage-1  rhaos-4.2-rhel-7-candidate  ocp-build/buildvm.openshift.eng.bos.redhat.com
```

It looks like when we build metadata containers for different environments, the number portion of the release value doesn't accumulate.

Also comparing the component name seems to be unnecessary.

sosiouxme: we are looking for the latest metadata in the stream (env - dev,stage,prod) that matches the exact operator build.
metadata does get rebuilt multiple times for the same operator build and stream; and when it does, the release gets bumped. the previous comparison was stringwise. i'm just changing it to numeric. I don't actually anticipate it being anything other than an incrementing number, or being greater than 9 for that matter, but if i can be robust against those edge cases, i want to.
you have a point about comparing the component though. the brew command should never return anything that's not the right component.

doozerlib/operator_metadata.py
sosiouxme: seemed conceivable to me that some operator would have `-container` in its name.
`my-super-container-manager-operator-container`
this could probably use an assertion that there was actually something to replace, though.

PR #152, opened 2019-12-05
Get rid of unicode warnings

thiagoalessio:
```
/usr/lib/python2.7/site-packages/doozerlib/cli/__main__.py:126: Warning:
Click detected the use of the unicode_literals __future__ import.  This
is heavily discouraged because it can introduce subtle bugs in your
code.  You should instead use explicit u"" literals for your unicode
strings.  For more information see
https://click.palletsprojects.com/python3/
```

### DIFF
--- a/doozerlib/__init__.py
+++ b/doozerlib/__init__.py
@@ -1,4 +1,3 @@
-from __future__ import unicode_literals
 from .runtime import Runtime
 from .pushd import Dir
 
--- a/doozerlib/assertion.py
+++ b/doozerlib/assertion.py
@@ -5,7 +5,6 @@ the asserted condition is not met.
 The use of the FileNotFound exception makes this Python3 ready.
 Making them functions keeps the exception definition localized.
 """
-from __future__ import unicode_literals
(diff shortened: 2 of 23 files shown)


vfreex:
It is controversial of the use `unicode_literals`. See discussions on [Should I import unicode_literals?](https://python-future.org/unicode_literals.html)
From my opinion, if our goal is migrating to Python 3 and deprecating Python 2 eventually, then changing the signature of our functions and always using unicode is not much a problem and can probably help us discover issues earlier (since Python 2 doesn't distinguish `byte` and `str`, much code may never call `encode`/`decode` correspondingly).
Hence, I would be in favor of disabling the warning rather than following it.

thiagoalessio:
@vfreex that was an attempt to solve `sriov-network-operator` build failures (error msg had to do with unicode) ... but Justin found out the root cause of that, therefore we can disregard this PR for now.

vfreex:
@thiagoalessio Alright. Anyway really appreciated your investigation.

PR #153, opened 2019-12-06
Force unicode run/decode

jupierce:


jupierce:
fyi @thiagoalessio (but not until you are back from PTO!)

doozerlib/cli/__main__.py
```diff
@@ -2192,7 +2192,17 @@ def update_operator_metadata(runtime, nvr_list, stream, image_ref_mode, merge_br
         # merge_branch = runtime.group_config.operator_metadata_branch
 
     try:
-        ok = ThreadPool(cpu_count()).map(operator_metadata.update_and_build, [
+
+        def gotta_catchem_all(*args, **kwargs):
```
sosiouxme: :rofl:

sosiouxme: but we really do need this

doozerlib/operator_metadata.py
```diff
@@ -229,7 +229,8 @@ def rename_csv_yaml_filename_of_arch_copy(self, arch):
             '{}.0.'.format(self.channel),          # e.g. 4.2.0
             '{}.0-{}.'.format(self.channel, arch)  # e.g. 4.2.0-s390x
         )
-        exectools.cmd_assert('mv {} {}'.format(original_filename, new_filename))
+        if original_filename != new_filename:
```
sosiouxme: hm. ok that'll work but... i'd rather just rip out the renaming

PR #154, opened 2019-12-10
make dotconfig respect unicode templates

adammhaile:
@thiagoalessio This is pretty urgent as CPaaS is dead in the water on doozer support until it's fixed and pushed to PyPi
Because the template was now `unicode` the type check was failing and and empty template was written.


thiagoalessio:
Hi @adammhaile , your fix is present in version `0.5.36` https://pypi.org/project/rh-doozer/0.5.36/

PR #156, opened 2019-12-20
Add use_source_version option for cpaas users

adammhaile:
@thiagoalessio @sosiouxme @tbielawa 
This is for cpaas compatibility. Basically many groups outside of OpenShift control their image version by including it in the github dockerfile. As it currently stands doing this is impossible as you either override the version with --version on the CLI or it uses what's in distgit.
What I added is simple: if `use_source_version: true` is in the image config it will fetch the version from the "upstream" Dockerfile after it has been copied into the distgit dir but before the modifications are made (this was the least intrusive). Since I'm just updating the `version` var in `rebase_images` the rest flows from there as normal.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1794,6 +1794,12 @@ class ImageDistGitRepo(DistGitRepo):
             if self.has_source():
                 self._merge_source()
 
+                # before mods, check if upstream source version should be used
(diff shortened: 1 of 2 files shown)


adammhaile:
Thanks!

sosiouxme:
v0.5.37

PR #157, opened 2019-12-29
Improve command executions

vfreex:
1. Make `Runtime._parallel_exec` interruptible so it won't hang forever if it expects user input.
2. When cloning a git repo but user input is required, fail the command rather than hang forever.

Also fix broken unit test.

### DIFF
--- a/doozerlib/constants.py
+++ b/doozerlib/constants.py
@@ -1,0 +1,8 @@
+from __future__ import absolute_import, print_function, unicode_literals
+
+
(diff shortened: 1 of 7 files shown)


sosiouxme:
@vfreex this seems like a good idea, can you rebase?

vfreex:
@sosiouxme rebased and fixed the broken unit test.

tbielawa:
@vfreex this looks good, probably ready to merge
/lgtm

PR #158, opened 2019-12-29
ART-128: Support ODCS composes for OSBS builds

vfreex:
This PR gives Doozer the ability to start OSBS builds with ODCS composes instead of repofiles.

Usage
-----

This PR doesn't enable ODCS mode by default. To building an image with ODCS mode, you must follow those steps:

1. Add `odcs` section to ocp-build-data image config.

If you know the list of RPMs that the image will install, you can use `manual` mode:

``` yaml
odcs:
  packages:
    list:
    - kibana
    mode: manual
```

Otherwise, You can use `auto` mode. `auto` mode will use the list of packages from source `container.yaml` or make all packages associated to the Koji build target available:

```yaml
odcs:
  packages:
    mode: auto
```

YUM repos defined in `enabled_repos` section will also be available as well.

Images without `odcs` section will not be built in ODCS mode. To make all images without `odcs` section in a build group assume ODCS auto mode, you can turn on the `odcs_aggressive` feature gate in ocp-build-data group config:

```yaml
doozer_feature_gates:
  # odcs_aggressive: true to build an image with ODCS auto mode even if its image config doesn't have `odcs` field defined
  odcs_aggressive: true
```

2. Run `images:rebase` or `images:update-dockerfile` with `--odcs-mode` option:

```bash
doozer --odcs-mode --group 'openshift-3.11' -i logging-kibana5 images:rebase --version=v3.11.161 --release=888 -m 'bump version to 3.11.161-888' --push
doozer --odcs-mode --group 'openshift-3.11' -i logging-kibana5 images:update-dockerfile --version=v3.11.161 --release=888 -m 'bump version to 3.11.161-888' --push
```

If you don't want to specify `--odcs-mode` option explicitly, You can add the following the content to ocp-build-data group config:

```yaml
doozer_feature_gates:
  odcs_enabled: true # whether to enable Doozer ODCS support (without explicitly specifying --odcs-mode option)
```

Doozer will generate and write ODCS related configs to image's distgit repo.

3. Build the image with `images:build`. If the image enables ODCS, Doozer will be aware of the ODCS config from the dist git repo and will not inject repofiles.

Behavior changes
----------------
This is a list of important changes to the existing ODCS support:

1. The old `auto` mode tries to guess what RPM packages the image requires. It does so by matching and parsing the `yum install` commands from Dockerfile then filter the result with `repoquery`. I believe this is error-prone because text-based processing instead of really executing the commands can't get a full list packages, and it doesn't handle RPM dependencies. The new `auto` mode will firstly look at container.yaml provided by upstream source. If it provides such a RPM list, we will use that list. Otherwise we will make all RPM packages within the current Koji build target available.

2. The old implementation has an implicit `pre` mode, which means if upstream source provides a list of packages in container.yaml, Doozer will use that list regardless of the mode in the `odcs` section. I don't think having such a mode is a good choice because it just ignores whatever packages listed in `odcs` section of our ocp-build-data image config. I removed that mode. If the image owner wants to use upstream source provided list, he should use the new `auto` mode.

3. The old implementation will require an `--odcs` option when running `images:build` in order to bypass the repofile injection. I think it is redundant because whether ODCS is enabled can be determined by looking at the container.yaml in distgit repo. I simply removed the `--odcs` option and simply reuse the `repo-type` option from non-ODCS mode for specifying whether signed packages are required.

4. Precedence change: `odcs` and `arches` sections in image config > `container_yaml` section in image config > existing container.yaml in source repo.


doozerlib/distgit.py
```diff
@@ -941,20 +902,32 @@ def _build_container(
...
         else:
             if repo_type:
                 repo_list = list(repo_list)  # In case we get a tuple
                 repo_list.append(self.metadata.cgit_url(".oit/" + repo_type + ".repo"))
 
-            if repo_list:
-                # rhpkg supports --repo-url [URL [URL ...]]
-                cmd_list.append("--repo-url")
-                cmd_list.extend(repo_list)
+        if repo_list:  # Current OSBS versions support the combination of ODCS composes with repository files.
```
sosiouxme: Was not the case before, but I see in https://mojo.redhat.com/docs/DOC-1159997#jive_content_id_Can_I_use_repourls_parameter_with_ODCS_composes that's changed...

However I don't think we want to take advantage of this. Everything we're shipping should be in pulp repos, except our own candidate tag, and for that we no longer want to point at "latest" -- that's kind of the whole point here, to use an ODCS compose for that.

doozerlib/cli/__main__.py
```diff
@@ -2262,9 +2262,6 @@ def operator_metadata_latest_build(runtime, nvr, stream, operator_list):
 
 def main():
     try:
-        if 'REQUESTS_CA_BUNDLE' not in os.environ:
-            os.environ['REQUESTS_CA_BUNDLE'] = '/etc/pki/tls/certs/ca-bundle.crt'
```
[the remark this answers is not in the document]
vfreex: This part of change was included because I was mostly working with this PR and another (python 3 migration and cleanup) at the same time then I involved some changes from there and didn't drop them. I'll drop  this kind of change.

BTW, I remembered I fixed that issue (but forgot to tell everyone) on buildvm by reinstalling `python-certifi` RPM some day.  The RPM version uses system provided CAs rather than vendored CAs from PyPI.

doozerlib/distgit.py
vfreex: Do you think it is better to remove this feature when ODCS mode is enabled?

vfreex: I don't know if we ever specify a `--repo-list` to `doozer images:build` during runtime (looks like at lease we don't use that option in `aos-cd-jobs`). I reverted this part of change. If some day we do have the requirement of specifying runtime repo url even when ODCS is enabled, we can add it back.

vfreex:
Code updated.

vfreex:
@sosiouxme @adammhaile @shiywang PTAL.

vfreex:
PR rebased.

doozerlib/distgit.py
```diff
@@ -331,146 +331,110 @@ def image_build_method(self):
...
+        image_specific_arches = set(self.metadata.config.arches)
+        if not image_specific_arches:
+            # assume all global arches if no image specific arches are defined
+            arches = list(global_arches)
         else:
-            source_container_yaml = {}
+            # only include arches from metadata that are valid globally
+            missing_arches = image_specific_arches - global_arches
+            if missing_arches:
+                raise ValueError("Image specific arches ({}) are not enabled in group.yml.".format(", ".join(missing_arches)))
```
sosiouxme: note this now raises an error if you put an arch in an image that isn't in the group, where it would have silently ignored it before. i think this is the correct approach (and ISTR there's something else catching that use case anyway), just want that to be visible to anyone looking back on this PR.

that said, it _would_ be useful in some cases to go back to silent ignorance. this comes up when we disable an arch in the group.yml and don't want to have to go around and remove it from the individual images as well (especially if we plan to re-enable it shortly).

doozerlib/distgit.py
```diff
@@ -331,146 +331,110 @@ def image_build_method(self):
...
-                res = list(set(res))
-                if res:
-                    packages.extend(res)
-
-            if odcs.packages.list:
-                config.compose.packages.extend(odcs.packages.list)
-
-            if odcs.packages.exclude:
-                exclude = set(odcs.packages.exclude)
+            if config["compose"].get("packages") is list:
```
sosiouxme: shouldn't this be detecting if it is an instance of `list`, not if it `is list`?
```suggestion
            if isinstance(config["compose"].get("packages"), list):
```
unit tests would help :)

doozerlib/distgit.py
```diff
@@ -331,146 +331,110 @@ def image_build_method(self):
...
+                # container.yaml with packages was given from source
+                self.logger.info("Use ODCS package list from source")
             else:
-                exclude = set([])
-
-            config.compose.packages.extend(packages)
-            config.compose.packages = list(set(config.compose.packages) - exclude)  # ensure unique list
+                config["compose"]["packages"] = []  # empty list donates all packages from the current Koji target
```
sosiouxme:
```suggestion
                config["compose"]["packages"] = []  # empty list composes all packages from current Koji target
```

doozerlib/distgit.py
```diff
@@ -941,11 +904,23 @@ def _build_container(
...
+        osbs_image_config_path = os.path.join(self.distgit_dir, "container.yaml")
+        if os.path.isfile(osbs_image_config_path):
+            with open(osbs_image_config_path, "r") as f:
+                image_config = yaml.safe_load(f)
+            odcs_enabled = "compose" in image_config
+
+        if odcs_enabled:
+            self.logger.info("About to build image in ODCS mode.")
+            if repo_type == 'signed':
+                signing_intent = 'release'
```
sosiouxme: there could be a use case for allowing beta here... deal with it if it comes up i guess.

doozerlib/image.py
```diff
@@ -90,85 +53,6 @@ def base_only(self):
         """
         return self.config.base_only
 
-    def get_rpm_install_list(self, valid_pkg_list=None):
-        """Parse dockerfile and find any RPMs that are being installed
```
sosiouxme: i am so glad to see this go :smile:

PR #159, opened 2019-12-30
Fully migrate to Python 3 and cleanup some dead code

vfreex:
The code changes were initially generated by `futurize --stage2`,
then I carefully reviewed all code changes and manually fixed all issues I noticed to pass the unit test.
    
This PR will also bump the version to `1.0.0-rc.1` (Wouldn't it be nice to have version >= 1?)
and prompt a warning message like pip if Doozer is running with Python 2:
    
```sh
$ python2 ./doozer --version
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of Doozer will drop support for Python 2.7.
Doozer v1.0.0-rc.1
Python v2.7.17 (default, Oct 21 2019, 17:20:57)
[GCC 9.2.1 20190827 (Red Hat 9.2.1-1)]

The Doozers don't mind their buildings being eaten;
if the Fraggles didn't eat the constructions,
the Doozers would run out of building space,
and if they ran out of building space,
they would have to move away from Fraggle Rock
or else they would die.
```


vfreex:
Removed WIP mark.

I also found an issue with `tag_exists` function in `metadata.py`. It is already broken because it accesses url like `http://registry-proxy.engineering.redhat.com/v1/repositories/openshift3/ose-ansible/tags/v3.11.161-668` to check if a tag exists. But that API actually returns `HTTP 403 Forbidden` error and that function mistakenly treats that error as `tag doesn't exist`.

Authenticating with buildvm's keytab doesn't help with the HTTP 403 error.

It breaks after applying the porting change because with the new urllib API, that function will raise an exception.

@sosiouxme @jupierce Can we remove the `tag_exists` check from `build_container` or this is something easy to fix?

vfreex:
PR rebased.

vfreex:
PR rebased

vfreex:
After merging this PR, build container will not work due to broken tag_exists check which was silently ignored with the old urllib API. Using this PR https://github.com/openshift/doozer/pull/165 to fix that check.

vfreex:
Rebased, again.

vfreex:
Rebased

sosiouxme:
@vfreex if you have a free minute could you resolve these conflicts? they mostly look easy (stuff deleted due to odcs) but i'd feel better if you did it. i may take a crack at it later though :)

sosiouxme:
I resolved the conflicts (hopefully correctly) and this is a good week to break things, so I will go ahead and merge this to try it out.

PR #160, opened 2020-01-02
Add functional tests

vfreex:
Add functional test suite for Doozer. The test suite works by invoking `doozer` as a subprocess then checking for the exit code and output.

Currently the test suite doesn't mock any services, such as dist-git, Brew, etc, but interacts with real production services. This is not perfect but can significantly help our Python 3 migration.

Some test functions are not currently implemented but only a stub with `FIXME` marked. This is because those Doozer commands are either hard to test or they are already broken.

See `functional_tests/README.md` for more information and known issues.


vfreex:
Created https://github.com/openshift/doozer/pull/161 to test build commands.

functional_tests/test_images.py
```diff
@@ -0,0 +1,138 @@
...
+        ])
+        working_dir = get_working_dir()
+        repo_path = os.path.join(working_dir, "distgits", "containers", "aos3-installation")
+        repo = pygit2.Repository(repo_path)
+        ref = repo.references["HEAD"].resolve()  # type: pygit2.Reference
+        git_commit = ref.peel(pygit2.Commit)  # type: pygit2.Commit
+        self.assertIn("test revert", git_commit.message)
+
+    def test_images_push_distgit(self):
+        # FIXME: images:push-distgit is not tested because we don't have an option to dry run
```
sosiouxme: we could push to a private branch (any branch beginning with `private-` which is allowed to create on demand) and then nuke the branch afterward.

functional_tests/test_images.py
```diff
@@ -0,0 +1,138 @@
...
+        pass
+
+    def test_images_print_config_template(self):
+        # FIXME: images:print-config-template is not tested because it is broken
+        pass
+
+    def test_images_query_rpm_version(self):
+        # FIXME: images:query-rpm-version is not tested because it is broken
```
sosiouxme: hm... really? we used it all the time in the refresh-images job up until it was deleted...

PR #165, opened 2020-01-10
fix broken tag_exists

vfreex:
The current `tag_exists` function is broken due to brew-pulp to registry-proxy migration.
It was not noticed because HTTP 403 is treated as `tag doesn't exist`.

After porting this function to Python 3, it will raise an `urllib.error.HTTPError` which prevents all image builds from being started.

Fixed this function and added unit tests.


vfreex:
@joepvd  Thanks.

PR #167, opened 2020-01-14
[ART-1451] Use UTF-8 in operator_metadata file operations

thiagoalessio:
A follow up on https://github.com/openshift/doozer/pull/164/files because 4.4 appregistry is still broken due to encoding issues on metadata YAMLs.


thiagoalessio:
Thanks a lot @sosiouxme !

PR #168, opened 2020-01-14
Tolerate when tito init does not make commit

jupierce:


doozerlib/runtime.py
```diff
@@ -868,7 +868,6 @@ def resolve_source_head(self, parent, meta):
         returns SHA of head.
         :param parent: Name of parent requesting the source
         :param meta: The MetaData object to resolve source for
-        :param required: Whether an error should be thrown or None returned if it cannot be determined
```
sosiouxme: :man_facepalming: oops

PR #174, opened 2020-02-04
opmeta: don't build opmetas multi-arch

sosiouxme:
Operator metadata containers will never be executed. They are just
glorified tarballs; there is no need to build all of the available
architectures and risk something going wrong with one randomly.

### DIFF
--- a/doozerlib/operator_metadata.py
+++ b/doozerlib/operator_metadata.py
@@ -89,6 +89,7 @@ class OperatorMetadataBuilder:
         self.update_current_csv_shasums()
         self.merge_streams_on_top_level_package_yaml()
         self.create_metadata_dockerfile()
+        self.create_container_yaml()
         return self.commit_and_push_metadata_repo()
 
     @log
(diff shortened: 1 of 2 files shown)


vfreex:
@sosiouxme PR needs rebase to include latest py3 test.
Also `get_builtin_module` was removed from the latest test code. You can use `io.open()` then mock `io` module.

doozerlib/operator_metadata.py
```diff
@@ -267,6 +268,18 @@ def create_metadata_dockerfile(self):
             self.stream)
         del(metadata_dockerfile.labels['release'])
 
+    def create_container_yaml(self):
+        """Use container.yaml to disable unnecessary multiarch"""
+        filename = "{}/{}/container.yaml".format(self.working_dir, self.metadata_repo)
+        with open(filename, "w") as cyml:
```
vfreex: Use `io.open()` instead of builtin `open()` until we drop support for Python 2.7. And we can also mock `io` module to make test easier.

tests/test_operator_metadata.py
```diff
@@ -587,6 +587,24 @@ def test_create_metadata_dockerfile(self):
         shutil.rmtree('/tmp/my-operator')
         shutil.rmtree('/tmp/my-dev-operator-metadata')
 
+    def test_create_container_yaml(self):
+        container_yaml_filename = '/my/working/dir/my-operator-metadata/container.yaml'
+        writer = flexmock(write=None)
+
+        mock = flexmock(get_builtin_module())
```
vfreex: `get_builtin_module` was removed from the latest test code.

thiagoalessio:
Should I bring this over to [ART-1728](https://issues.redhat.com/browse/ART-1728) ?

sosiouxme:
@thiagoalessio please. we only need a single arch on bundles.

sosiouxme:
obsolete

PR #175, opened 2020-02-04
metadata: ART-1520 handle 403 response

sosiouxme:
https://issues.redhat.com/browse/ART-1520


vfreex:
Merged. CI test failed, but I believe it was not caused by this PR.

PR #176, opened 2020-02-05
Speed up images:print when dockerfile parsing is not required

jupierce:


tbielawa:
@jupierce I pulled your branch down, ran the tests and saw the failure above. And then I did `git pull --rebase origin master` and ran the tests again and it passed. Maybe you should too.

PR #177, opened 2020-02-05
[ART-1513] Retry tag query with backoff

joepvd:


tbielawa:
py27 error is due to use of `nonlocal` keyword.

PR #178, opened 2020-02-10
images:print: don't crash if no builds found for an image

vfreex:
In 4.4 we have some new images. `images:print` crashes when there is an image without any builds.
This breaks our appregistry job.

This PR will catch the `no builds detected` error to make `images:print` proceed. Note NVR will be `name--` when this happens but I don't know how to handle that better.

See https://issues.redhat.com/browse/ART-1532


joepvd:
Think this behavior needs to differ between a GA and z-stream release. Think a crash during z-stream is desirable, whereas for GA (or.. maybe pre-`rc`), more leniency is apt.

joepvd:
/lgtm

A _Nice to have_ is not a hard requirement. :shipit:

PR #180, opened 2020-02-13
Add image build profile support for custom Brew target and so on

vfreex:
A build profile is set of configuration values which allows us to build our project using different configurations.
This feature is added to Doozer to support custom Brew target (especially for image build with ODCS) and build RPMs for RHEL versions using a single build command.

To define build profiles, we should add the following section to ocp-build-data group.yml:

```yaml
build_profiles:
  image:
    unsigned:
      targets:
      - rhaos-{MAJOR}.{MINOR}-rhel-7-containers-candidate
      signing_intent: unsigned
      repo_type: unsigned
      repo_list: []
    signed:
      targets:
      - rhaos-{MAJOR}.{MINOR}-rhel-7-signed-containers-candidate
      signing_intent: release
      repo_type: signed
      repo_list: []
  rpm:
    default:
      targets:
      - rhaos-{MAJOR}.{MINOR}-rhel-7
      - rhaos-{MAJOR}.{MINOR}-rhel-8

default_image_build_profile: unsigned
default_rpm_build_profile: default
```

This PR only covers the image portion.

When building an image, use the `--profile` to specify the build profile you want to use:

``` bash
doozer --debug -g openshift-3.11 --latest-parent-version --images logging-kibana5 --profile=signed images:build
```

The above command will use the profile `signed`. If `--profile` is not given, it will use the `default_image_build_profile` value in group.yml. If `default_image_build_profile` is not defined as well, this feature will be disabled to keep compatibility with the old behavior.

The `--repo-type` option and `repo_list` option will always override the build profile values. This is also to maintain backward compatibility.

Signed-off-by: Yuxiang Zhu <yuxzhu@redhat.com>

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -128,6 +128,7 @@ def print_version(ctx, param, value):
               help="Path to rhpkg config file to use instead of system default")
 @click.option("--brew-tag", metavar="BREW_TAG",
               help="Override brew tag to expect for images built")
+@click.option("--profile", metavar="NAME", default="", help="Name of build profile")
 @click.pass_context
 def cli(ctx, **kwargs):
     global CTX_GLOBAL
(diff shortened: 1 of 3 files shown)


vfreex:
This PR https://github.com/openshift/ocp-build-data/pull/308 defines build profile for 3.11.

vfreex:
Looks like a py3-only keyword `nonlocal` was introduced to master. Should I modify that part of code to get py27 test passed, or disable py27 test eventually?

sosiouxme:
I'm asking our only other user if it's ok to drop support for py 2.7. Hopefully it's not a problem.
With that in mind I think I lean toward disabling/ignoring the test.
I'm going to ignore it to merge this :)

PR #181, opened 2020-02-17
[ART-1311] Drop Python2 support

thiagoalessio:


vfreex:
Would also like to remove Python 2.7 info from setup.py. otherwise LGTM.

vfreex:
The unit test failure is not related to this PR, which is broken on master. https://github.com/openshift/doozer/pull/180 has the unit test fix

thiagoalessio:
@vfreex setup.py done :heavy_check_mark:

vfreex:
LGTM, but needs to wait for @sosiouxme 's confirmation with other users:

> @sosiouxme: I'm asking our only other user if it's ok to drop support for py 2.7. Hopefully it's not a problem.


/lgtm
/hold

sosiouxme:
/hold cancel
we're good, let's do this.

PR #182, opened 2020-02-18
Bring tag_exists check back

vfreex:
https://github.com/openshift/doozer/pull/180 disabled `tag_exists` check. Needs to enable it.


vfreex:
It just reverts to the old existing code, I think it is ok to merge :)

PR #184, opened 2020-02-20
ART-1536: scan-sources should not count disabled images as changed

vfreex:
This bug happens because `mode: disabled` and `mode: wip` only disable autoloading that image.yml when `--images` is not specified, but it will be loaded anyway if another image defines the disabled image as a dependent.

This PR https://github.com/openshift/ocp-build-data/pull/304 has removed `openshift-enterprise-template-service-broker-operator` from the dependent list of `template-service-broker`.

To prevent this issue from happening again, we can skip disabled images when scanning sources unless `--load-disabled` is specified.

Signed-off-by: Yuxiang Zhu <yuxzhu@redhat.com>

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -405,10 +405,11 @@ def config_scan_source_changes(runtime, as_yaml):
     Print a list of configs that were scanned and whether the source differs from the distgit.
     """
     _fix_runtime_mode(runtime)
-    CONFIG_RUNTIME_OPTS['disabled'] = False  # unless explicitly indicated, ignore disabled configs
     runtime.initialize(**CONFIG_RUNTIME_OPTS)
     results = dict(rpms=[], images=[])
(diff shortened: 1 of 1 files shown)


doozerlib/cli/__main__.py
```diff
@@ -397,8 +397,9 @@ def images_update_dockerfile(runtime, stream, version, release, repo_type, messa
 
 @cli.command("config:scan-sources", short_help="Determine if source repo differs from distgit content")
 @click.option("--yaml", "as_yaml", default=False, is_flag=True, help='Print results in a yaml block')
+@click.option("--include-disabled", default=False, is_flag=True, help="Including disabled images")
```
sosiouxme: There is an option like this on the runtime:
```
--load-disabled              Treat disabled images/rpms as if they were enabled
```

I think I would prefer to just reuse that here.

vfreex: Makes sense. I've changed the code to respect that option.

PR #190, opened 2020-02-28
Prevent rpmcfg from creating non-origin commit

jupierce:
In the previous implementation, it was possible to have an rpm build create a commit in a global source alias without pushing it back to origin. If an image build later used that source directory for building an image, it would get encode a commit_sha that had no basis in origin.
The new implementation ensures that:
1. if you create a commit, you must push it. 
2. if two rpms are building out of the same source alias, they must wait for one another (to prevent competing modifications / commits).

### DIFF
--- a/doozerlib/rpmcfg.py
+++ b/doozerlib/rpmcfg.py
@@ -113,6 +113,13 @@ class RPMMetadata(Metadata):
     def commit_changes(self, scratch):
         if not self.tag:
             raise ValueError('Must run set_nvr() before calling!')
+
+        # If not using tito, you can operate in different modes:
+        #  do_push_commit: push new commit and tag to origin
+        do_push_commit = self.config.content.build.push_release_commit and not scratch
+        #  do_push_tag: only push a tag
+        do_push_tag = self.config.content.build.push_release_tag and not scratch
+
         with Dir(self.source_path):
             if self.config.content.build.use_source_tito_config:
                 # just use the tito tagger to change spec and tag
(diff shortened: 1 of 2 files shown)


doozerlib/rpmcfg.py
```diff
@@ -123,16 +130,22 @@ def commit_changes(self, scratch):
...
-                name=self.config.name, version=self.version, release=self.release
-            )
-            exectools.cmd_assert(['git', 'commit', '-m', commit_msg])
+                # If you are using source, tito, you are going to want your tags/commits in origin
+                exectools.cmd_assert("git push origin")
+
+            else:
+                if do_push_tag:
+                    exectools.cmd_assert(["git", "tag", "-am", "Release with doozer", self.tag])
+                    exectools.cmd_assert(f"git push origin {self.tag}")
```
sosiouxme: doesn't look like we did this before... odd? but seems harmless.

doozerlib/rpmcfg.py
```diff
@@ -384,52 +397,53 @@ def build_rpm(self, version, release, terminate_event, scratch=False, retries=3,
...
-        }
 
-        try:
-            def wait(n):
-                self.logger.info("Async error in rpm build thread [attempt #{}]: {}".format(n + 1, self.qualified_name))
-                # Brew does not handle an immediate retry correctly, wait
-                # before trying another build, terminating if interrupted.
-                if terminate_event.wait(timeout=5 * 60):
-                    raise KeyboardInterrupt()
+        with self.runtime.get_dir_lock(self.source_path):
```
sosiouxme: don't think we have an RPMs building from the same source, only images. still, i suppose it could theoretically happen.

PR #192, opened 2020-03-03
[ART-1240] Clean out env vars in dist-git only Dockerfiles

jupierce:


doozerlib/distgit.py
```diff
@@ -1396,6 +1391,19 @@ def update_distgit_dir(self, version, release, prev_release=None):
...
+                # necessary for us to clean up our older env injections if this is a dist-git only
+                # Dockerfile.
+                if line.lower().startswith('env'):
+                    skip_line = False
+                    for var_name in env_vars:
+                        if f' {var_name}=' in line:
+                            skip_line = True
+                            break
+                    if skip_line:
+                        continue
```
sosiouxme: Define a variable to break from outer loop, yuk... How about
```suggestion
                    if any(f' {var_name}=' in line for var_name in env_vars):
                        continue
```

doozerlib/distgit.py
```diff
@@ -1409,6 +1417,12 @@ def update_distgit_dir(self, version, release, prev_release=None):
             with io.open('Dockerfile', 'w', encoding="utf-8") as df:
                 for comment in oit_comments:
                     df.write("%s %s\n" % (OIT_COMMENT_PREFIX, comment))
+
+                # specify new env vars in the dockerfile
+                if env_vars:
+                    env_line = "ENV " + " ".join([key + "=" + val for key, val in env_vars.items()])
+                    df.write(f'{env_line}\n')
```
sosiouxme: i'm missing something about how this is being done right now (something is injecting env vars, i WROTE it, just don't see it at the moment) but this does not seem right. you would have to write it at the beginning of each stage, not just once.

jupierce: New push injects after each FROM

PR #193, opened 2020-03-05
Restore + to KUBE_GIT_MINOR

jupierce:
Historically, a '+' was added to the kube minor version. Some version parsing broke when this was removed in the move away from custom tito tagging.


sosiouxme:
doozer updated on buildvm

PR #196, opened 2020-04-01
ART-1723: Handle failed Brew tasks with build records left

vfreex:
Double check the build record and only ignore the error if the build state is complete.


sosiouxme:
@joepvd just for context i have seen builds succeed and then the task fails while either tagging the build or uploading logs to the server.

we don't always actually know beforehand what the NVR will be -- in 3.11 at least we let the R auto-increment.

doozerlib/constants.py
```diff
@@ -6,3 +6,5 @@
     "GIT_SSH_COMMAND": "ssh -oBatchMode=yes",
     "GIT_TERMINAL_PROMPT": "0",
 }
+
+BREW_HUB = "https://brewhub.engineering.redhat.com/brewhub"
```
sosiouxme, line 10: we have this configured in group.yml, could it use that please?

vfreex:
@joepvd 

The original title of this PR was mistakenly messed up and it was misleading. I've updated the the title and hope it is more clear.

> This means that if the build is `COMPLETE` all is ok, regardless of the state of of the `task`.
> Unsure what `task` is doing after `build` is done, but it does not deviate from the current intent.

The added code path will only be executed when the state of task is failed, such as https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=27571644. Currently Doozer will ignore the error if the error message is like `BuildError: Build for ose-cluster-image-registry-operator-container-v4.3.10-202003310434 already exists, id 1153049` because it thinks the tasks failure is cause by a duplicate NVR. However we already see such an exception like this task the build state is not complete: https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=1153049. This PR adds an additional check on the build state so that Doozer will only ignore when such an error occurs and build state is complete.

> 
> I think these changes are an improvement over the current state. But it could be better:
> 
> now:
> 
> ```
> task1 fails, build1 succeeds
> marked as failure, retry

No sure what you mean by build1 succeeds here. We haven't seen this situation happens and doozer never checks the state of build1 currently.

> task2 fails, nvr already exists
> doozer concludes: all is good as nvr exists
> ```
exactly.

To summarize, the current logic is:

- task1 succeeds -> doozer marks as success
- task1 fails with `build already exists` -> doozer marks as success
- task1 fails with other errors -> doozer marks as failure

> 
> This pr:
> 
> ```
> task1 fails, build1 succeeds
> marked as failure, retry
> task2 fails, nvr already exists
> doozer checks if build for nvr succeeded, and concludes: all is good as nvr exists
> ```

I think you misunderstand the Brew concept here. In brew's domain, build and nvr are 1-1 mapping and you can almost think they are the same concept. You can't have 2 builds with the same nvr because a build can be indexed by a nvr.

This PR further checks the state of the build (or nvr) and only concludes all is good if and only if the nvr exists and its state is complete.

> 
> Proposal:
> 
> ```
> task1 fails, build1 succeeds
> upon task failure, doozer checks if the build1 succeeded
> and does not retry
> ```
> 
> This would mean to not parse the error, but to look if a build with the expected nvr exists and is marked as `COMPLETE`.

doozerlib/constants.py
vfreex: Sure. updated.

PR #197, opened 2020-04-01
Allow build arches to be overridden on the command line

jupierce:
We need group.yml to be a source of truth for arches when opening Cincinnati PRs in order to make the determination of whether to add a `+amd64` suffix for images. For example:
1. We are building 4.2.99 for arches [x86_64, s390x]
2. 4.3 is still GA only for a single arch [x86_64]
3. When we open a Cincinnati PR to put 4.2.99 in the 4.3 channels, we need to detect that 4.3 is not yet shipping for s390x and add `4.2.99+amd64` 

The OTA team has asked us to do this out of an abundance of caution. If we do not, then when a 4.3.z ships _with_ s390x support, it will immediately allow 4.2.99 s390x installations to upgrade to the new 4.3.z. This isn't necessarily a bad thing, but if we prevent it with by only adding `4.2.99+amd64` to the channels, it decreases the test surface that OTA needs to concern itself with.

So, on to the purpose of this PR. For various reasons, ART has been asked in the past to temporarily turn on CPU architectures in the builds. This means that we have to, for example, alter the arches field in the 4.3 group.yml temporarily to get a 4.3 s390x nightly out to our test groups. 

During this interval, 4.3's group.yml cannot be used as a source of truth for opening 4.2 Cincinnati PRs (again, we need to know what arches are shipping for 4.3 in order to determine which +arch CPU suffix to use, if any). This PR allows us to maintain truth in group.yml while still building test builds for other arches.

You can accomplish this two different ways. 
1. First, you can now provide a set of architectures on the doozer command line via `--arches` . Jobs can be altered to allow this from Jenkins parameters.
2. `arches` can be left to reflect GA status in group.yml  while `arches_override` (also in group.yml) can be used to temporarily override the GA arches for test purposes.

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -32,6 +32,7 @@ from multiprocessing.pool import ThreadPool
 from multiprocessing import cpu_count
 from dockerfile_parse import DockerfileParser
 from doozerlib import dotconfig
+from doozerlib import gitdata
 
 click.disable_unicode_literals_warning = True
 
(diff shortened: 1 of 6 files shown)


sosiouxme:
still seems like a good idea to have this. but practically speaking, it would only be useful if we also made it such that when group.yml had
```
arches: (or arch_override)
  - x86_64
```

and the individual image had:

```
arches: (or arch_override)
  - x86_64
  - s390x
```

... it took the intersection of that and did not try to build s390x (currently the commits for disabling s390x have to also remove it where mentioned in the images).

jupierce:
Intersection logic has now been implemented.

PR #199, opened 2020-04-10
Remove tag_exists check before starting a Brew image build

vfreex:
For the background, see discussions at
https://coreos.slack.com/archives/GDBRP5YJH/p1586479678450600?thread_ts=1586474171.448900&cid=GDBRP5YJH.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -755,59 +755,55 @@ class ImageDistGitRepo(DistGitRepo):
         target_image = ":".join((self.org_image_name, target_tag))
 
         try:
-            if not self.runtime.local and not scratch and self.org_release is not None \
-                    and self.metadata.tag_exists(target_tag):
-                self.logger.info("Image already built for: {}".format(target_image))
-            else:
-                # If this image is FROM another group member, we need to wait on that group member
-                # Use .get('from',None) since from is a reserved word.
-                image_from = Model(self.config.get('from', None))
-                if image_from.member is not Missing:
-                    self._set_wait_for(image_from.member, terminate_event)
-                for builder in image_from.get('builder', []):
-                    if 'member' in builder:
-                        self._set_wait_for(builder['member'], terminate_event)
+            # If this image is FROM another group member, we need to wait on that group member
+            # Use .get('from',None) since from is a reserved word.
+            image_from = Model(self.config.get('from', None))
+            if image_from.member is not Missing:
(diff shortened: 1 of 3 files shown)


doozerlib/distgit.py
```diff
@@ -736,59 +736,55 @@ def build_container(
...
+                        raise KeyboardInterrupt()
+
+                if len(profile["targets"]) > 1:
+                    # FIXME: Currently we don't really suppport building images against multiple targets,
+                    # or we would overwrite the image tag when pushing to the registry.
+                    # `targets` is defined as an array just because we want to keep consistency with RPM build.
+                    raise DoozerFatalError("Building images against multiple targets is not currently supported.")
+                target = '' if not profile["targets"] else profile["targets"][0]
+                exectools.retry(
+                    retries=retries, wait_f=wait,
```
sosiouxme, line 784: i think we still want this, to the extent that anyone might use `doozer --local` - we don't want to retry local builds.

doozerlib/distgit.py
```diff
@@ -736,59 +736,55 @@ def build_container(
...
-                            raise KeyboardInterrupt()
-
-                    if len(profile["targets"]) > 1:
-                        # FIXME: Currently we don't really suppport building images against multiple targets,
-                        # or we would overwrite the image tag when pushing to the registry.
-                        # `targets` is defined as an array just because we want to keep consistency with RPM build.
-                        raise DoozerFatalError("Building images against multiple targets is not currently supported.")
-                    target = '' if not profile["targets"] else profile["targets"][0]
-                    exectools.retry(
-                        retries=(1 if self.runtime.local else retries), wait_f=wait,
```
sosiouxme, line 788: @vfreex Sorry, should have attached my previous comment to the "this" that I wanted to keep.
To the extent that anyone might use `doozer --local` - we don't want to retry local builds.

vfreex: @sosiouxme This is dead code and local build doesn't reach here. It is in the `else` branch of `if self.runtime.local:` https://github.com/openshift/doozer/pull/199/files#diff-8fa829dca0dd48a4369166e4f848476eL757

vfreex:
Rebased.

doozerlib/distgit.py
sosiouxme: I see. There probably shouldn't be an `else:` at all since that branch returns; but OK.

PR #201, opened 2020-04-10
hack day: doozer bucket list

jupierce:
1. Added a --cache-dir options to doozer. If specified, doozer will populate distgit & sources repos there and use those repos as clone --reference. All doozer instances, regardless of group, can use this same directory. Each run will update the cached repos. This reduces images:clone time from about 20m to 3m.

2. Implemented a timing analysis tool to see where time was being spent. After a doozer run, execute something like `./doozer analyze:debug-log ../wd/debug.log > analysis.txt' . This will show you what each thread was doing over the course of the doozer run. 

3. I ran the analysis tool on a run on `images:rebase`. Not unexpectedly, the lack of threading stuck out like relative to the now fast cloning. The final commit adds threading to rebase and update-dockerfile. This forced several changes in the way file I/O was handed in these methods to make them threadsafe.

All together, changes 1-3 can reduce the rebase time from > 1 hour to ~10 minutes. 

4. The `Fix update-dockerfile KUBE_ env loss` commit is to address https://issues.redhat.com/browse/ART-1730

5. The `jenkins/brew information to distgit labels` commit aims to give us easy to reference information about brew tasks & jenkins jobs. Want to know what the last failure was for 4.5 cli image? Browse to http://pkgs.devel.redhat.com/cgit/containers/openshift-enterprise-cli/log/?h=rhaos-4.5-rhel-7 and Ctrl-F for `doozer-failure` . The annotated tag will provide you information about the Jenkins job and brew task that failed.

~~6. The `Force rebuild when ocp-build-data changes` will report source change when ocp-build-data changes. In someone changes their configuration or we update group.yml, we should rebuild instead of waiting 24 hours.~~ DROPPED

PR #202, opened 2020-04-10
Hack'n'Hustle: Support rebasing images from specified Git commit hashes

vfreex:
This PR adds a new option `--upstream` to the `images:rebase` command so that Doozer can build images from specified Git commits.
This would give us the ability to rebuilding an image any time without introducing unwanted code changes.

To specify the commit hashes, use this syntax:

```bash
doozer images:rebase --upstream distgit_key_1=commitish_1 --upstream
distgit_key_2=commitish_2 ...
```

For example:

```bash
doozer -g openshift-4.4 --images ose-metering-ansible-operator:c0d9aaf5a901cc9b40068cac0b9c037579d72dc7,openshift-enterprise-ansible-operator \
    --latest-parent-version images:rebase -m 'test' --version v4.4 --release 202001010000 --upstream ose-metering-ansible-operator=c0d9aaf5a901cc9b40068cac0b9c037579d72dc7
```

Note you images specified in `--images` without `--upstream` will use
the tip commit of the branch so it will not break the backward
compatibility.

Limitations:
- Building RPMs from commit hashes is not covered in this PR.
- The behavior of `--latest-parent-version` is not changed, so it could use a base image from the latest Brew build.
- commit-ish format is not checked, so it relies on `git` to check that.

In the future we can also have an option like `--parent` with the same
syntax to allow us to
specify parent image NVRs.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1745,6 +1745,10 @@ class ImageDistGitRepo(DistGitRepo):
         self.env_vars_from_source = {}
 
         with Dir(self.source_path()):
+            if self.metadata.commitish:
+                self.runtime.logger.info(f"Rebasing image {self.name} from speicified commit-ish {self.metadata.commitish}...")
+                cmd = ["git", "checkout", self.metadata.commitish]
+                exectools.cmd_assert(cmd)
             # gather source repo short sha for audit trail
             rc, out, _ = exectools.cmd_gather(["git", "rev-parse", "--short", "HEAD"])
             self.source_sha = out.strip()
--- a/doozerlib/runtime.py
+++ b/doozerlib/runtime.py
@@ -249,7 +249,8 @@ class Runtime(object):
     def initialize(self, mode='images', clone_distgits=True,
                    validate_content_sets=False,
                    no_group=False, clone_source=True, disabled=None,
-                   config_excludes=None):
+                   config_excludes=None,
+                   upstream_commitish_overrides={}):
 
         if self.initialized:
             return
@@ -411,6 +412,9 @@ class Runtime(object):
             image_ex = list(exclude_keys)
             rpm_ex = list(exclude_keys)
             image_keys = flatten_list(self.images)
+
+            self.upstream_commitish_overrides = upstream_commitish_overrides
+
             rpm_keys = flatten_list(self.rpms)
 
             filter_func = None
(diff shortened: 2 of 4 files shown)


doozerlib/distgit.py
```diff
@@ -1601,6 +1601,10 @@ def _merge_source(self):
         """
 
         with Dir(self.source_path()):
+            if self.metadata.commitish:
+                self.runtime.logger.info(f"Rebasing image {self.name} from speicified commit-ish {self.metadata.commitish}...")
```
sosiouxme, line 1605:
```suggestion
                self.runtime.logger.info(f"Rebasing image {self.name} from specified commit-ish {self.metadata.commitish}...")
```

doozerlib/runtime.py
```diff
@@ -398,7 +398,19 @@ def filter_disabled(n, d):
...
+            self.pinned_commitishes = {}
+            # `image` can either be a `distgit key` or `distgit_key:commit-ish`
+            for image in flatten_list(self.images):
+                image_split = image.split(":")
+                distgit_key = image_split[0]
+                image_keys.append(distgit_key)
+                if len(image_split) < 2:
+                    continue
+                commitish = image_split[1]
+                # TODO: verify `commitish` format
```
sosiouxme, line 411: meh... I think it's fine not to.

sosiouxme:
BTW we need this for RPMs too. Not as important, but I think we'll have a use case.

vfreex:
@sosiouxme RPM will be covered... Though I have not come up with a plan to rebuild images after that RPM rebuilds... Maybe we will link RPMs to images in the fuy

doozerlib/runtime.py
vfreex: Well, now I think we'd better not to implement that check on Doozer. Doozer should be able to build from anywhere. We can implement the check in the component invoking Doozer, like a Jenkins job.

jupierce:
Mainly a UI observation: pairing the commit with the -i argument makes me think you are specifying a distgit commit. The -i entries are effectively distgit identifiers.
`--upstream github.com....:hash` would be clearer to me. If a distgit has an upstream that matches a specified upstream, it would use that hash instead of the default.

vfreex:
@jupierce Makes sense. I will update later.

vfreex:
@jupierce I updated the code with following changes:

- I moved the hash specification to dedicated `--upstream` option using syntax `--upstream some-image=hash`. But it still uses distgit names as the identifiers. One consideration is we share sources for some images using "alias". If we override the hash when a distgit's upstream matches  "github.com...", it will override all images sharing that source. Another thought is that in the near future we may provide the input via Jenkins build parameter. It could be easier for people to use the distgit names everywhere, otherwise people have to lookup ocp-build-data for the source URLs in order to fill the parameter. Considering the option is named to `upstream`, I think it becomes more clear than reuse the `--images` option, and gives us the flexibility to only override the commits for a subset of images.
- https://github.com/openshift/doozer/pull/201 introduced a behavior change that shallow clone is used for better clone performance. I added an additional step to fetch the commit explicitly after git clone.

PTAL. Thanks.

vfreex:
@jupierce @sosiouxme Can we merge this or you guys still have suggestions?

PR #205, opened 2020-04-18
Fix issues with #200: Improve gen-payload logging and performance

vfreex:
Fixed the following issues with #200 :
1. in `__main__.py`: incorrect use of `lstrip`.
2. in of the `list_archives_by_builds` function, forgot to `continue` the loop if build ID is `None`, so the order of some images is screwed.

Also added some unit tests and additional checks on top of it.

I am doing a complete test with all images in 4.1 (single arch) and multi-arch (4.5) against the old behavior and will put the test result later.

CC @sosiouxme @jupierce PTAL

/hold

### DIFF
--- a/doozerlib/brew.py
+++ b/doozerlib/brew.py
@@ -112,8 +112,11 @@ def get_latest_builds(tag_component_tuples: List[Tuple[str, str]], session: koji
     tasks = []
     with session.multicall(strict=True) as m:
         for tag, component_name in tag_component_tuples:
+            if not (tag and component_name):
+                tasks.append(None)
+                continue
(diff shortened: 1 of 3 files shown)


vfreex:
OK, now we can ensure the results are the same:
```
[dev@ec21599b6258 doozer]$ diff -u ./old-4.5/ ./new-4.5
[dev@ec21599b6258 doozer]$ diff -u ./old-4.1/ ./new-4.1
[dev@ec21599b6258 doozer]$ 
```

[test_results.zip](https://github.com/openshift/doozer/files/4495876/test_results.zip)

jupierce:
Can we put UUID into the tag name? I can't think of any reason we would ever want to update a tag's sha.

PR #206, opened 2020-04-20
[ART-1159] allow non-semver versions so we can omit the .z

sosiouxme:
I'll take http://post-office.corp.redhat.com/archives/aos-team-art/2020-April/msg00109.html as approval that we can version the brew builds however we see fit.

This allows us to build packages and images with version `v4.4` and not have a confusing `.0` added on. However note that env vars aimed at build scripts in spec files and dockerfiles will continue to be set as if the version implicitly had `0` for any missing segments.

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -189,11 +189,9 @@ def cli(ctx, **kwargs):
     return ctx
 
 
-def validate_semver_major_minor_patch(ctx, param, version):
+def validate_version(ctx, param, version):
     """
-    For non-None, non-auto values, ensures that the incoming parameter meets the criteria vX.Y.Z or X.Y.Z.
-    If minor or patch is not supplied, the value is modified to possess
-    minor.major to meet semver requirements.
+    For non-None, non-auto values, ensures that first 3 version segments are integers (with optional v prefix)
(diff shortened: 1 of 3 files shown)


doozerlib/cli/__main__.py
```diff
@@ -205,15 +203,12 @@ def validate_semver_major_minor_patch(ctx, param, version):
     vsplit = version.split(".")
     try:
         int(vsplit[0].lstrip('v'))
-        minor_version = int('0' if len(vsplit) < 2 else vsplit[1])
-        patch_version = int('0' if len(vsplit) < 3 else vsplit[2])
+        int('0' if len(vsplit) < 2 else vsplit[1])
```
vfreex, line 206: This will allow minor version to be omitted. Is this intended?

sosiouxme: It seems unlikely we'll use that. Just thought some other user of doozer might appreciate it someday.

vrutkovs:
This breaks CVO, which expects version to be semver (trimming "v") - https://openshift-release.svc.ci.openshift.org/releasestream/4.4.0-0.nightly/release/4.4.0-0.nightly-2020-04-21-080252

PR #210, opened 2020-04-24
Re-enabled doozer tagging

jupierce:
https://bugzilla.redhat.com/show_bug.cgi?id=1827706 needs to be fixed for all releases before re-enabling: https://bugzilla.redhat.com/show_bug.cgi?id=1827706#c8

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -982,8 +982,7 @@ class ImageDistGitRepo(DistGitRepo):
                            '-m', f'{build_state} from {self.runtime.uuid}'
                            ]
                 tag_cmd.extend(messages)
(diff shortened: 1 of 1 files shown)


thiagoalessio:
Not merging yet, because previous comment says we should wait https://bugzilla.redhat.com/show_bug.cgi?id=1827706 fix to re-enable, and BZ is still in `NEW` status.

thiagoalessio:
@jupierce maybe we can merge it now? aforementioned BZ was marked as a duplicate of https://bugzilla.redhat.com/show_bug.cgi?id=1829028, which is now in `VERIFIED` state.

jupierce:
@thiagoalessio unfortunately, there is more upstream work before this could merge. In good news, this is completely unnecessary if we go down the simpledb route, so I have not been pushing on getting the upstream stuff fixed yet.

jupierce:
Obsolete.

PR #211, opened 2020-04-28
Generate source revisions for images

vfreex:
This PR adds a new command to generate source revisions for images.

- On the build the product day, run this command to record the upstream
  commit hashes as source revision objects for all images.
- Store the source revision objects to the apiserver using `oc` or
  `kubectl`.

Example 1: Generate source revisions for OpenShift 4.4

        doozer -g openshift-4.4 beta:releaseapi gen-source-revisions -r openshift-4.4.1 -o openshift-4.4.1.source-revisions.yaml

Example 2: Generate source revisions for specified components

        doozer -g openshift-4.4 -i ose-installer-artifacts,openshift-enterprise-cli beta:releaseapi gen-source-revisions -r openshift-4.4.1 -o openshift-4.4.1.source-revisions.yaml

Example YAML output:

```yaml
---
apiVersion: art.openshift.io/v1alpha1
kind: SourceRevision
metadata:
  labels:
    distgit-key: openshift-enterprise-cli
    image-name: ose-cli
    kind: image
    release: openshift-4.4.1
    release-stream: openshift-4.4
  name: openshift-4.4.1-image-ose-cli
spec:
  commitHash: 2576e482bf003e34e67ba3d69edcf5d411cfd6f3
  componentName: openshift-4.4-image-ose-cli
---
apiVersion: art.openshift.io/v1alpha1
kind: SourceRevision
metadata:
  labels:
    distgit-key: ose-installer-artifacts
    image-name: ose-installer-artifacts
    kind: image
    release: openshift-4.4.1
    release-stream: openshift-4.4
  name: openshift-4.4.1-image-ose-installer-artifacts
spec:
  commitHash: 78b817ceb7657f81176bbe182cc6efc73004c841
  componentName: openshift-4.4-image-ose-installer-artifacts
```

## Note
- The `openshift-release` package is a auto generated from the OpenAPI derived from the Golang structs: https://github.com/vfreex/release-apiserver/tree/master/python-openshift-release

### DIFF
--- a/doozerlib/runtime.py
+++ b/doozerlib/runtime.py
@@ -884,6 +884,45 @@ class Runtime(object):
         cmd.append(target_dir)
         exectools.cmd_assert(cmd, retries=3, on_retry=["rm", "-rf", target_dir], set_env=set_env)
 
+    def lookup_source(self, meta):
+        """
+        Looks up a source alias and returns source_details dict containing
+        that source. It doesn't clone any source.
+        NOTE: This function is a copy of `resolve_source` but doesn't really clone anything
+        """
+        source = meta.config.content.source
+
+        if not source:
+            return None
+
+        parent = f'{meta.namespace}_{meta.name}'
+
+        # This allows passing `--source <distgit_key> path` to
+        # override any source to something local without it
+        # having been configured for an alias
+        if self.local and meta.distgit_key in self.source_paths:
+            source['alias'] = meta.distgit_key
+            if 'git' in source:
+                del source['git']
+
+        source_details = None
+        if 'git' in source:
+            git_url = urllib.parse.urlparse(source.git.url)
+            name = os.path.splitext(os.path.basename(git_url.path))[0]
+            alias = '{}_{}'.format(parent, name)
+            source_details = dict(source.git)
+        elif 'alias' in source:
+            alias = source.alias
+        else:
+            raise DoozerFatalError('Error while processing source for {}'.format(parent))
+
+        if not source_details:  # old style alias was given
+            if self.group_config.sources is Missing or alias not in self.group_config.sources:
+                raise DoozerFatalError("Source alias not found in specified sources or in the current group: %s" % alias)
+            source_details = self.group_config.sources[alias]
+
+        return source_details
+
     def resolve_source(self, meta):
         """
         Looks up a source alias and returns a path to the directory containing
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@ dockerfile-parse >= 0.0.13
 future
 koji
 mock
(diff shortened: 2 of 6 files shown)


doozerlib/runtime.py
```diff
@@ -859,6 +859,45 @@ def git_clone(self, remote_url, target_dir, gitargs=None, set_env=None, timeout=
         cmd.append(target_dir)
         exectools.cmd_assert(cmd, retries=3, on_retry=["rm", "-rf", target_dir], set_env=set_env)
 
+    def lookup_source(self, meta):
```
sosiouxme, line 862: I think `detect_remote_source_branch` already did this pretty much?

Anyway, I want to throw a curveball here. I would expect we would do this after running (or at least initiating) the builds for a release. By that point, it's at least possible that the sources might already have drifted from what we built. Would we not want to capture what we built, rather than what is in the sources at some slightly later time?

We could do that by looking at the Dockerfiles in dist-git (via cgit). `doozer config:scan-sources` uses this logic already for comparing github sources with distgit contents.

requirements.txt
```diff
@@ -4,8 +4,10 @@ dockerfile-parse >= 0.0.13
 future
 koji
 mock
+openshift-release >= 0.0.2
 pyyaml >= 5.1
 retry
 requests
 requests_kerberos
+ruamel.yaml >= 0.15
```
sosiouxme, line 12: pyyaml wasn't good enough?

vfreex: It is more like a leftover when I exploring more format control over the YAML dumping. Seems it is not required anymore.

doozerlib/runtime.py
vfreex: @sosiouxme There were a lot of code changes since this PR was open especially after Justin merged the fast rebase change. Last time when I visited the doozer code, it is not simple to reuse the existing functions and I didn't have enough confidence to modify them. Now I think I need to revisit that part of code and see if any parts can be reused.

vfreex: Capturing the actual builds is not covered in this PR. Capturing commit hashes after the initial build doesn't seem provide more value than recording commit hashes first and it will produces 2 different code paths.
I think the following can be a transitional route before we get a smart trigger:
- prepare-release job: records commit hashes of branch tips for all components. Triggers custom build job for them.
- Patch-release job: bump the commit hash to latest branch tip for an image, or specific a custom commit. Triggers custom build for any changes. 
- custom build job: will always build from the last recorded commit. Users usually don't need to directly trigger this job anyway unless explicitly rebuild without code change.
- build-sync job will always create an image stream with builds that matches the recorded hashes
@sosiouxme @jupierce would like to hear suggestions from you guys.

PR #217, opened 2020-05-29
ART-1861: Set private_fix boolean for private build

vfreex:
Determine if the source contains private fixes by checking if the
private org branch commit exists in the public org, set `private_fix`
to True when there are private fixes.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -103,6 +103,8 @@ class DistGitRepo(object):
         self.source_latest_tag = None
         self.source_date_epoch = None
         self.source_url = None
+        # This will be set to True if the source contains embargoed (private) CVE fixes. Defaulting to False for distgit only repos.
+        self.private_fix = False
 
         # If we are rebasing, this map can be populated with
         # variables acquired from the source path.
@@ -1789,8 +1791,8 @@ class ImageDistGitRepo(DistGitRepo):
         # Initialize env_vars_from source.
         # update_distgit_dir makes a distinction between None and {}
         self.env_vars_from_source = {}
-
-        with Dir(self.source_path()):
+        source_dir = self.source_path()
+        with Dir(source_dir):
             if self.metadata.commitish:
                 self.runtime.logger.info(f"Rebasing image {self.name} from speicified commit-ish {self.metadata.commitish}...")
                 cmd = ["git", "checkout", self.metadata.commitish]
@@ -1798,16 +1800,20 @@ class ImageDistGitRepo(DistGitRepo):
             # gather source repo short sha for audit trail
             rc, out, _ = exectools.cmd_gather(["git", "rev-parse", "--short", "HEAD"])
             self.source_sha = out.strip()
-            rc, out, _ = exectools.cmd_gather(["git", "rev-parse", "HEAD"])
+            out, _ = exectools.cmd_assert(["git", "rev-parse", "HEAD"])
             self.source_full_sha = out.strip()
             rc, out, _ = exectools.cmd_gather("git log -1 --format=%ct")
             self.source_date_epoch = out.strip()
             rc, out, _ = exectools.cmd_gather("git describe --always --tags HEAD")
             self.source_latest_tag = out.strip()
 
-            rc, out, _ = exectools.cmd_gather(["git", "remote", "get-url", "origin"])
+            out, _ = exectools.cmd_assert(["git", "remote", "get-url", "origin"])
             out = out.strip()
-            self.source_url = self.runtime.get_public_upstream(out)  # Point to public upstream if there are private components to the URL
+            self.source_url, _ = self.runtime.get_public_upstream(out)  # Point to public upstream if there are private components to the URL
+
+            # Determine if the source contains private fixes by checking if the private org branch commit exists in the public org
+            if self.metadata.public_upstream_branch:
+                self.private_fix = not util.is_commit_in_public_upstream(self.source_full_sha, self.metadata.public_upstream_branch, source_dir)
 
             # If this is a go project, parse the Godeps for points of interest
             godeps_file = pathlib.Path(self.source_path(), 'Godeps', 'Godeps.json')
@@ -2028,6 +2034,10 @@ class ImageDistGitRepo(DistGitRepo):
 
             if self.config.content.source.modifications is not Missing:
                 self._run_modifications()
+
+        # TODO: adjust Doozer's execution for private build
+        if self.private_fix:
+            self.logger.warning("The source of this image contains embargoed fixes.")
 
         (real_version, real_release) = self.update_distgit_dir(version, release, prev_release)
 
--- a/doozerlib/runtime.py
+++ b/doozerlib/runtime.py
@@ -23,6 +23,7 @@ import signal
 import io
 import pathlib
 import koji
+from typing import Optional
 
 from doozerlib import gitdata
 from . import logutil
@@ -794,7 +795,7 @@ class Runtime(object):
 
         return self.streams[stream_name]
 
-    def get_public_upstream(self, remote_git):
+    def get_public_upstream(self, remote_git: str) -> (str, Optional[str]):
         """
         Some upstream repo are private in order to allow CVE workflows. While we
         may want to build from a private upstream, we don't necessarily want to confuse
@@ -805,7 +806,9 @@ class Runtime(object):
         into a new string with the public prefix in its place. If there is not
         applicable mapping, the incoming url will still be normalized into https.
         :param remote_git: The URL to analyze for private repo patterns.
-        :return: An https normalized remote address with private repo information replaced.
+        :return: tuple (url, branch)
+            - url: An https normalized remote address with private repo information replaced.
+            - branch: Optional public branch name if the public upstream source use a different branch name from the private upstream.
         """
         remote_https = util.convert_remote_git_to_https(remote_git)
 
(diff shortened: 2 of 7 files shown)


vfreex:
The CI test failure doesn't seem related to this PR.

doozerlib/util.py
```diff
@@ -96,6 +98,15 @@ def convert_remote_git_to_https(source):
     return re.sub(string=url, pattern=r'\.git$', repl='').rstrip('/')
 
 
+def is_commit_in_remote_repo(revision: str, repo_url: str):
+    rc, out, err = exectools.cmd_gather(["git", "fetch", "--", repo_url, revision])
```
jupierce, line 102: Does this work?

I tried this.. `a4f3878a68097a07d78f117886310517a5581dce` is the current HEAD in my doozer clone.
```
# Run test to see if this hash exists in openshift/origin. It doesn't, but..
$ git fetch -- https://github.com/openshift/origin a4f3878a68097a07d78f117886310517a5581dce
From https://github.com/openshift/origin
 * branch            a4f3878a68097a07d78f117886310517a5581dce -> FETCH_HEAD
# ^ No error
```

jupierce: The way I'm familiar with doing this is to:
1. add the public repo as a remote of the source repository (e.g. `git remote add public_upstream ...`)
2. `fetch public_upstream [branch]`  . This should be fast as our private clone should contain most of these public_upstream commits.
3. `git branch -r --containers <commit_hash>`  . If the public_upstream/[branch] is returned in this output, our current commit is public.

vfreex: @jupierce Weird, I got the following error when running your command. git version difference?

``` bash
git fetch -- https://github.com/openshift/origin a4f3878a68097a07d78f117886310517a5581dce
fatal: remote error: upload-pack: not our ref a4f3878a68097a07d78f117886310517a5581dce
fatal: The remote end hung up unexpectedly
```
Maybe I should change to your proposed way if that is not reliable.

vfreex: @jupierce Well, I think `git` may have some short circuit logic if the commit already exists in the local repo. Then it should be more reliable to use the `fetch` and `branch --contains` way.

vfreex: Seems adding `--depth=1` to `fetch` will force the commit is fetch from the remote URL. I will make an update here. If we still want to fetch the full history of public upstream (that will make your previous shallow clone optimization void unfortunately), I can make that change next week.

jupierce: With --cache-dir, I don't believe there is an appreciable speed up with --depth=1, so it won't be a loss for normal pipeline runs. But it will impact runs for individuals not using cache. 

Is the `--depth=1` fetch behavior documented? Otherwise it seems like we are exploiting a codepath that might be optimized in the future (like the non-depth call).

vfreex: I don't think `--depth=1` fetch is officially documented anywhere. I just found the usage in examples at StackOverflow. Seems we should go with the full fetch way to do that.

sosiouxme: If we can rely on `--depth=1` then I'm +1 on this.

I was trying this out and want to document what seemed like a weird thing to me. I was testing this out using https://github.com/kube-reporting/presto as the public source and https://github.com/sosiouxme/presto/ as the "private" fork (of course, it's not actually private).

So I added a commit https://github.com/sosiouxme/presto/commit/6f308005db1ca0f8aa3eac88b8999edc60444f2e and ran the rebase to see if it would be identified as an "embargoed fix". It wasn't because... as soon as I made that commit in my fork, it became visible in the original: https://github.com/kube-reporting/presto/commit/6f308005db1ca0f8aa3eac88b8999edc60444f2e (note, not on any branch, while it is on `release-4.5` in my fork).

Seems to be a github feature. I _really_ hope that wouldn't work with an actually-private repo (I don't think they're registered as forks). But let's be sure we test...

vfreex: I don't think --depth=1 fetch is officially documented anywhere. I just found the usage in examples at StackOverflow.
Seems we can use an empty dir only for fetching purpose. Full fetch with `git branch --contains` seems more expensive since we need a lock to ensure multithreaded fetch doesn't run out of order. Let me try and see which way is eaiser.

vfreex: @sosiouxme Seems it is more complicate than what we thought before. Now I am considering if a strict `git merge-base is-ancestor` check is needed.
I will try all the approaches and test if they will work.

vfreex: @sosiouxme GitHub doesn't allow us to fork a repo then make it private. We have to create a private repo then duplicate all the commits. I just tested if the duplicated repo is a not a fork, fetching the original will not work. But now I am hesitate if we should rely on this behavior.

Doing a full fork then `git branch --contains` could be a more reliable way. Actually I prefer to use `git merge-base is-ancestor` since we can limit the branch in the original repo, however currently we don't have the public upstream branch configured in ocp-build-data and I am not sure if we should do that...

vfreex: Anyway, I am currently exploring using the full fetch and `git branch --contains` way. The tricky part is locking. Let me read more about the code context and determine how to do that.

vfreex:
@jupierce @sosiouxme Here are the changes:
- Fetch public upstream source in `resolve_source()`. This process is protected by the source dir lock.
- Disable shallow clone when `--cache-dir` is used so that the public upstream fetch will be able to reuse commits in private upstream source.
- Determine if the current commit hash exists in public upstream source with `git merge-base --is-ancestor <public-upstream-branch> <current-commit>` when rebasing dist-git.
- openshift/ose has different branch names from openshift/origin. Add a new configuration field in ocp-build-data: https://github.com/openshift/ocp-build-data/pull/487/files

doozerlib/util.py
```diff
@@ -96,6 +98,34 @@ def convert_remote_git_to_https(source):
...
+    # rc, out, err = exectools.cmd_gather(["git", "-C", source_dir, "branch", "--remotes", "--list", "public_upstream/*", "--contains", revision])
+    # if rc == 0:
+    #     return out.strip() != ""
+    # if "no such commit" in err:
+    #     return False
+    cmd = ["git", "merge-base", "--is-ancestor", "public_upstream/" + public_upstream_branch, "--", revision]
+    rc, out, err = exectools.cmd_gather(cmd)
+    if rc == 0:
+        return True
+    if err == "" or "Not a valid commit name" in err:
```
jupierce, line 124: Same caution on using stderr messages. Does `--contains` not work as expected?

vfreex: `--contains` works, but it doesn't set exit code. It seems to me `merge-base --is-ancestor` is more reliable to distinguish `commit doesn't exist in a branch` and `general command error`.

doozerlib/runtime.py
```diff
@@ -942,24 +942,29 @@ def resolve_source(self, meta):
...
-
             self.logger.info("Attempting to checkout source '%s' branch %s in: %s" % (url, clone_branch, source_dir))
             try:
                 self.logger.info("Attempting to checkout source '%s' branch %s in: %s" % (url, clone_branch, source_dir))
                 # clone all branches as we must sometimes reference master /OWNERS for maintainer information
                 gitargs = [
-                    '--no-single-branch', '--branch', clone_branch, '--depth', '1'
+                    '--no-single-branch', '--branch', clone_branch
                 ]
+                if not self.cache_dir:
```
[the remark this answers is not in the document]
vfreex: It is for cloning the private upstream source. It just preserves the current behavior (though shallow clone doesn't mean much after we have full clone for public upstream source)

doozerlib/distgit.py
```diff
@@ -1789,26 +1791,30 @@ def _merge_source(self):
...
             rc, out, _ = exectools.cmd_gather("git describe --always --tags HEAD")
             self.source_latest_tag = out.strip()
 
-            rc, out, _ = exectools.cmd_gather(["git", "remote", "get-url", "origin"])
+            out, _ = exectools.cmd_assert(["git", "remote", "get-url", "origin"])
             out = out.strip()
             self.source_url = self.runtime.get_public_upstream(out)  # Point to public upstream if there are private components to the URL
 
+            # Determine if the source contains private fixes by checking if the private org branch commit exists in the public org
+            if self.metadata.public_upstream_branch:
```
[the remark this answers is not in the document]
vfreex: seems good

doozerlib/util.py
vfreex: What I want to achieve here is just to distinguish `commit doesn't exist` from other command execution errors. Another choice is using a library like [pygit2](https://github.com/libgit2/pygit2) to do that.

vfreex: @jupierce The doc https://git-scm.com/docs/git-merge-base#Documentation/git-merge-base.txt---is-ancestor said the exit code is 1 if the commit is not in the branch, while for other errors the exit code is not 1. It is clearly documented so I think we can use the exit code.

doozerlib/runtime.py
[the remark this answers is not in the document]
vfreex: I've updated the code to do a shallow clone only if public upstreams are not specified

doozerlib/distgit.py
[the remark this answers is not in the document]
vfreex: PR https://github.com/openshift/ocp-build-data/pull/487 has the proposed change for 4.5.

vfreex:
@jupierce Could you have another look?

vfreex:
Relevant ocp-build-data PRs:
- 4.6 https://github.com/openshift/ocp-build-data/pull/491
- 4.5 https://github.com/openshift/ocp-build-data/pull/487
- 4.4 https://github.com/openshift/ocp-build-data/pull/492
- 4.3 https://github.com/openshift/ocp-build-data/pull/493
- 4.2 https://github.com/openshift/ocp-build-data/pull/494
- 3.11 https://github.com/openshift/ocp-build-data/pull/495

doozerlib/util.py
jupierce: If `--contains` runs and exits with zero, you simply trust the content of stdout. If the stdout does not include the upstream branch, you know to set private_fix.

PR #222, opened 2020-06-08
ART-1879: Implement p0/p1 release convention for private fix builds

vfreex:
This PR implements the following items:

- When doozer detects that it is building an RPM with a private fix OR rebasing an image distgit with a private fix, it should enforce a release (as in NVR) convention. When a private fix is detected, add .p1 to the release component. When it is not, add .p0.
- When calling doozer images:rebase, the user/pipeline should also be able to specify "--embargoed". This should cause all images being rebased to include a .p1 field.
- doozer images:update-dockerfile should preserve the .p? field when updating the release.
- Also adds "--embargoed" option to `doozer rpms:build`.

Note:
- needs to revert e358ad760faebc66272153db62129e6b699a11f5 before merging this PR.
- Detecting if an image has private RPMs is not covered.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1448,7 +1448,7 @@ class ImageDistGitRepo(DistGitRepo):
                     if self.runtime.group_config.public_upstreams:
                         raise ValueError("Failed to bump the release: Neither 'release' is specified in the Dockerfile nor we can use OSBS auto-bumping when a public upstream mapping is defined in ocp-build-data.")
                     self.logger.info("No release label found in Dockerfile; bumping unnecessary -- osbs will automatically select unique release value at build time")
-                    release = prev_release
+                    release = None
 
             # If a release is specified, set it. If it is not specified, remove the field.
             # If osbs finds the field, unset, it will choose a value automatically. This is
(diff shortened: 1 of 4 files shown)


doozerlib/distgit.py
```diff
@@ -1426,35 +1426,47 @@ def update_distgit_dir(self, version, release, prev_release=None):
...
+                    if self.runtime.group_config.public_upstreams:
+                        release += ".p?"  # appended '.p?' field will be replaced with .p0 or .p1 later
                 else:
                     # When 'release' is not specified in the Dockerfile, OSBS will automatically
                     # find a valid value for each build. This means OSBS is effectively auto-bumping.
                     # This is better than us doing it, so let it.
+                    if self.runtime.group_config.public_upstreams:
+                        raise ValueError("Failed to bump the release: Neither 'release' is specified in the Dockerfile nor we can use OSBS auto-bumping when a public upstream mapping is defined in ocp-build-data.")
                     self.logger.info("No release label found in Dockerfile; bumping unnecessary -- osbs will automatically select unique release value at build time")
+                    release = prev_release
```
jupierce, line 1451: maybe?
```suggestion
                    release = None
```

doozerlib/distgit.py
```diff
@@ -1812,7 +1824,7 @@ def _merge_source(self):
             self.source_url, _ = self.runtime.get_public_upstream(out)  # Point to public upstream if there are private components to the URL
 
             # Determine if the source contains private fixes by checking if the private org branch commit exists in the public org
```
jupierce, line 1826:
```suggestion
            # If private_fix has not already been set (e.g. by --embargoed), determine if the source contains private fixes by checking if the private org branch commit exists in the public org
```

vfreex:
Rebased.

vfreex:
Rebased.

sosiouxme:
@vfreex @jupierce Ready to merge but one question: it looks like this requires a corresponding change in our jenkins jobs to start adding the `.p?` suffix - is that right? Could doozer just do that itself?

vfreex:
@sosiouxme The initial idea is auto adding `.p?` to the release value but @jupierce prefers to explicitly passing `.p?` to doozer's command line. So yes, we need this PR https://github.com/openshift/aos-cd-jobs/pull/2262 to change our Jenkins pipelines.

vfreex:
@jupierce @sosiouxme I'd like to merge this to test with next week's release. Let me know if you see anything breaks.

PR #223, opened 2020-06-11
ART-1908: Implement `doozer detect-embargo tag|nvr|pullspec|release` verbs

vfreex:
Implements `doozer detect-embargo tag|nvr|pullspec|release` verbs for finding embargoes in different kinds of build artifacts.

Also refactors the code to reuse most of the related logic in `releases:gen-payload` and adds more unit tests (because this is hard to test).

- Check whether one or more brew tags have an embargoed rpm or image:

        $ doozer --group=openshift-4.6 detect-embargo tag rhaos-4.6-rhel-8-candidate rhaos-4.6-rhel-7-candidate

- Check whether one or more brew tags have an embargoed rpm or image but exclude specified tags (you don't have to specified base tags with `--exclude` because shipped builds will be automatically excluded):

        $ doozer --group=openshift-4.6 detect-embargo tag rhaos-4.6-rhel-8-candidate rhaos-4.6-rhel-7-candidate --exclude rhaos-4.6-rhel-8 --exclude rhaos-4.6-rhel-7

- Check whether one or more nvrs have an embargo:

        $ doozer --group=openshift-4.6 detect-embargo nvr openshift-4.6.0-202007110420.p1.git.0.4de1d1d.el8 openshift-enterprise-hyperkube-container-v4.6.0-202007171513.p0 openshift-enterprise-tests-container-v4.6.0-202007172010.p1

- Check whether one or more arbitrary brew images have an embargo:

        $ doozer --group=openshift-4.6 detect-embargo pullspec registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-ironic@sha256:caa56dc45580ec5103564125cbb87f41559aec89de4a8efd099777463381180c

- Check whether one or more release payloads have an embargo:

        $ doozer --group=openshift-4.5 detect-embargo release quay.io/openshift-release-dev/ocp-release:4.5.2-x86_64

Use `--yaml` or `--json` to print the result as yaml or json document. If neither `--yaml` nor `--json` is given, it will exit with status `0` if true, or with status `2` if not. Errors are signaled by a non-zero status that is not `2`.


vfreex:
Rebased

doozerlib/cli/detect_embargo_cli.py
```diff
@@ -0,0 +1,175 @@
...
+        jobs = runtime.parallel_exec(lambda pullspec, terminate_event: get_nvr_by_pullspec(pullspec), pullspecs,
+                                     min(len(pullspecs), multiprocessing.cpu_count() * 4, 32))
+        nvrs = jobs.get()
+
+    if tags:
+        with runtime.shared_koji_client_session() as brew_session:
+            build_type = None if kind == "all" else kind
+            logger.info(f"Fetching builds from Brew tags {tags + exclude_tags}...")
+            tasks = []
+            with brew_session.multicall(strict=True) as m:
```
jupierce, line 95: This approach may be slightly faster by using a single multicall, but I would prefer readability here. e.g.
```python
def get_tag_nvrs(tag_event_list):  # just a local closure
    with brew_session... as m:
        for tag, event in [unpack_tag_event(t) for t in tag_event_list]:
     ....
included_nvrs = get_tag_nvrs(tags)
excluded_nvrs = get_tag_nvrs(excluded_tags)
```

jupierce:
I'd like to work on https://issues.redhat.com/browse/ART-1880 before revisiting this. I think some of the logic implemented in 1880 will be useful in making detect-embargo provide better answers than it could otherwise.

PR #224, opened 2020-06-15
Un-revert "ART-1861: Set private_fix boolean for private build"

vfreex:
This reverts commit e358ad760faebc66272153db62129e6b699a11f5.


vfreex:
This just reverts a previous commit and shouldn't break the CI.
That CI error is just because we are running out of space on buildvm.

PR #225, opened 2020-06-16
ART-1916: Fix content_sets cannot be empty array error

vfreex:
It seems recent Brew/OSBS has enforced a schema check for content_sets.yml,
which disallows an empty set to be declared:
https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/schemas/content_sets.json.

Let's omit empty content sets to avoid this error.

### DIFF
--- a/doozerlib/repos.py
+++ b/doozerlib/repos.py
@@ -274,13 +274,14 @@ class Repos(object):
 
         result = {}
         for a in self._arches:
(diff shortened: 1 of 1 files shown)


shiywang:
@vfreex have we already bumped into this kind of errors in our jobs?

PR #228, opened 2020-06-22
Fix private fix detection is not run for some images

vfreex:
There are 2 issues:
1. For dist-git only images, we can't detect if it includes private fixes because it has no upstream sources. This PR will try to preserve the private fix boolean from Dockerfile.
2. For images sharing a common source (using source alias), cache the public upstream resolution result of the first run to a map then retrieve it later.


vfreex:
Guys, I would like to merge this to unblock this week's build issue.

PR #230, opened 2020-06-24
ART-1962: Protect resolving public upstreams with lock

vfreex:
Per discussion https://coreos.slack.com/archives/GDBRP5YJH/p1592890124250700, use a lock to protect resolving public upstreams in order to simplify the logic and prevent potential race conditions:

- Replaces self.source_paths with self.source_resolutions to cache more info than source paths (like public upstream url and branch)
- Reuses the existing lock to protect the new field.

### DIFF
--- a/doozerlib/runtime.py
+++ b/doozerlib/runtime.py
@@ -4,6 +4,7 @@ standard_library.install_aliases()
 from future.utils import as_native_str
 from multiprocessing.dummy import Pool as ThreadPool
 from contextlib import contextmanager
+from collections import namedtuple
 
 import os
 import sys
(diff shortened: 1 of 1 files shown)


doozerlib/runtime.py
```diff
@@ -983,7 +1000,7 @@ def resolve_source(self, meta):
         # This allows passing `--source <distgit_key> path` to
         # override any source to something local without it
         # having been configured for an alias
-        if self.local and meta.distgit_key in self.source_paths:
+        if self.local and meta.distgit_key in self.source_resolutions:
```
jupierce, line 1003: As I look through this, it seems that at some point, resolve_source was designed to be single threaded. There are a number of race conditions I see now (introduced not by your PR, but by the use of these methods in a muli-threaded context). Your PR is subject to those pre-existing races. Soo.. what do you think about putting this item on the backlog until we can fully revisit threading in resolve_source?

vfreex: Sounds good to me. I'll log this to the backlog.

PR #231, opened 2020-06-24
Force user to specify workspace

jupierce:
A single doozer invocation that uses a /tmp workspace can exhaust memory on buildvm. Let's stay safe and start to require a workspace to be specified.

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -97,7 +97,7 @@ def print_version(ctx, param, value):
               expose_value=False, is_eager=True)
 @click.option("--data-path", metavar='PATH', default=None,
               help="Git repo or directory containing groups metadata")
-@click.option("--working-dir", metavar='PATH', default=None,
+@click.option("--working-dir", metavar='PATH', required=False,
               help="Existing directory in which file operations should be performed.\n Env var: DOOZER_WORKING_DIR")
 @click.option("--user", metavar='USERNAME', default=None,
               help="Username for rhpkg. Env var: DOOZER_USER")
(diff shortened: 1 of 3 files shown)


doozerlib/cli/__main__.py
```diff
@@ -97,7 +97,7 @@ def print_version(ctx, param, value):
               expose_value=False, is_eager=True)
 @click.option("--data-path", metavar='PATH', default=None,
               help="Git repo or directory containing groups metadata")
-@click.option("--working-dir", metavar='PATH', default=None,
+@click.option("--working-dir", metavar='PATH', required=True,
```
sosiouxme, line 100: This does not seem to allow using the env var or config file to specify this?

doozerlib/runtime.py
```diff
@@ -281,15 +274,9 @@ def initialize(self, mode='images', clone_distgits=True,
             click.echo("Group must be specified")
             exit(1)
 
-        if self.working_dir is None:
```
sosiouxme, line 284: it would probably be ok, though, to fail here if no working dir had been specified?

doozerlib/cli/__main__.py
jupierce: Should be good now.

vfreex:
How about moving the default working directory to `$HOME/.cache/doozer`?

jupierce:
@vfreex I like the idea from a usability perspective (I hate taking away the simplicity of using /tmp working dirs). But on buildvm:
1. If a non-jenkins user used this for anything sizable, it would exhaust `/`
2. If the user su'ed to jenkins, they would be placing this content on the NFS - which can be painfully slow (e.g. 40 minutes vs 10 for a full clone). 
I'm not sure what the right answer is.

vfreex:
@jupierce Where do you expect the user to use `--working-dir` on buildvm? Randomly choosing a working dir doesn't sound good either.
Ideally the default working directory will be cleaned up before Doozer exits, although it is not ensured.

Some other thoughts:
- Use `~/.cache/doozer` but check the free space with `shutil.disk_usage(path)` ahead and exit with error if free space is too little (e.g. < 5G)
- Use a subdirectory of the currently directory.

PR #233, opened 2020-06-26
Fix p? leak

jupierce:


jupierce:
@vfreex had to merge quick to clean up leaks of .p? into Dockerfiles and .specs. But please take a look.

PR #236, opened 2020-07-02
ART-1880: Modify gen-payload to target ocp priv namespaces

vfreex:
This PR is for supporting the new embargoed CVE handling workflow. See https://issues.redhat.com/browse/ART-1296 for more details.
    
- If public upstreams are not defined in ocp-build-data, no changes will happen.
- Excludes embargoed images from ocp[-arch] imagestreams, which will only include non-embargoed images.
- All images will be mirrored to ocp[-arch]-priv imagestreams.
- RHCOS will be mirrored from ocp[-arch] to ocp[-arch]-priv.

There are some details in the implementation that I am not very sure. Please have a look:) @jupierce @sosiouxme


doozerlib/cli/release_gen_payload.py
```diff
@@ -0,0 +1,334 @@
...
+        mirror_filename = 'src_dest.{}'.format(key)
+        imagestream_filename = 'image_stream.{}'.format(key)
+        target_is_name = is_name
+        target_is_namespace = is_namespace
+        if arch != 'x86_64':
+            target_is_name = '{}-{}'.format(target_is_name, arch)
+            target_is_namespace = '{}-{}'.format(target_is_namespace, arch)
+
+        if private:
+            target_is_namespace += "-priv"  # should we also append `-priv` suffix to target_is_name?
```
jupierce, line 159: No need to append. Imagestream names will be named with same within public & private namespaces.

doozerlib/cli/release_gen_payload.py
```diff
@@ -0,0 +1,334 @@
...
+                                                                                                  release=entry['release'],
+                                                                                                  arch_ext=arch_ext,
+                                                                                                  tag_name=tag_name,
+                                                                                                  uuid=runtime.uuid)
+
+        # Save the default SRC=DEST 'oc image mirror' input to a file for
+        # later.
+        with io.open(mirror_filename, 'w+', encoding="utf-8") as out_file:
+            for tag_name in mirroring[key]:
+                dest = build_dest_name(tag_name)
```
jupierce, line 178: I'd like the tags for our destinations to no longer include any identifying information - no dates, no runtime.uuid. I do not want an attacker to be able to derive a tag for an embargoed image based on a tag name for an unembargoed image. The easiest way I know how to do this to tag the image with a string including its own sha.
So instead of `quay.io/{orgrepo}:{version}-{release}{arch_ext}-ose-{tag_name}-{uuid}` , something like `quay.io/{orgrepo}:sha-{image_sha}`

doozerlib/cli/release_gen_payload.py
```diff
@@ -0,0 +1,334 @@
...
+                    'tags': tag_list,
+                }
+            }
+
+            for tag_name in mirroring[key]:
+                tag_list.append({
+                    'name': tag_name,
+                    'from': {
+                        'kind': 'DockerImage',
+                        'name': build_dest_name(tag_name)
```
jupierce, line 205: The `sha-{image_sha}` convention suggested above might best be incorporated into the mirroring structure.

doozerlib/cli/release_gen_payload.py
```diff
@@ -0,0 +1,334 @@
...
+                    }
+                })
+
+            if private:
+                # mirroring rhcos
```
jupierce, line 210: Well done, but let's comment this out for now. I need to explore some options with the rhcos team.

doozerlib/cli/release_gen_payload.py
```diff
@@ -0,0 +1,334 @@
...
+
+    # second, if an image's release field includes .p1, it is embargoed
+    embargoed_build_ids = {b["id"] for b in suspects if b["release"].endswith(".p1")}
+
+    # finally, look at the buildroots of .p0 images in case they include unshipped .p1 rpms
+    suspects = [b for b in suspects if b["id"] not in embargoed_build_ids]  # all .p0 images
+    suspect_build_ids = {b["id"] for b in suspects}
+    suspect_archives = [ar for archives in archives_list if archives for ar in archives if ar["build_id"] in suspect_build_ids]
+    logger.info(f'Fetching {len(suspect_archives)} buildroots for {len(suspects)} image builds...')
+    suspect_buildroots = brew.list_image_buildroots([ar["id"] for ar in suspect_archives], brew_session)
```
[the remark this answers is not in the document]
vfreex: I used that code as a reference so this is the same api call, but I misunderstood the concept of "buildroot".

doozerlib/cli/release_gen_payload.py
```diff
@@ -0,0 +1,334 @@
...
+    result = False
+    for index, tags in enumerate(tag_lists):
+        shipped = False
+        for tag in tags:
+            if tag["name"].endswith("-released"):  # a shipped RPM should have a tag like `RHBA-2020:2734-released`
+                shipped = True
+                break
+        rpm_ship_statuses[uncached[index]["build_id"]] = shipped
+        if not shipped:
+            result = True
```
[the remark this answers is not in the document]
vfreex: Here we want to cache more shipping statuses to reduce future Brew queries. It won't affect the return value.

vfreex:
@jupierce I've updated the PR. PTAL.

vfreex:
Added `--event-id` option.

sosiouxme:
@jupierce got a minute to re-review? even the single commit is tl;dr for me :-/

jupierce:
Looks about right to me. @vfreex can you work up the corresponding change to `buildSyncGenInputs`, `buildSyncMirrorImages`, and `buildSyncApplyImageStreams` in aos-cd-jobs? They currently only assume an arch suffix. I think these should just change to look for src_dest.* / image_stream.*.yaml .

vfreex:
@jupierce Sure. will create a PR for that change.

vfreex:
@jupierce 
I created https://github.com/openshift/aos-cd-jobs/pull/2280 for aos-cd-jobs changes.
I also added a new commit to this PR in order to exclude dummy image updates if `cli` is embargoed.

doozerlib/cli/release_gen_payload.py
```diff
@@ -225,6 +225,10 @@ def build_dest_name(tag_name):
                 if tag_name not in mirroring[key]:
 
                     if 'cli' not in mirroring[key]:
+                        if runtime.group_config.public_upstreams and not private:
+                            # Exclude any dummy image updates for public namespaces if cli is embargoed.
+                            # If "cli" image is also missing in *-priv namespaces, the error will be raised later when attempting to update a dummy image in *-priv namespaces.
+                            break
```
sosiouxme, line 231: This doesn't look like the right location for this logic - 
```
if 'cli' not in mirroring[key]:
```
is just a check to make sure there *is* a dummy `cli` image to use in the replacement. The code here will only have an effect if there isn't one.

I think, after confirming that there is a `cli` image, we want to check here whether the `cli` image is ok to sync, and if it's not, `continue` here (just omit `tag_name` entirely; usually this should leave the previous dummy image in place).

vfreex: Well, I think the logic here has become a little hard to read (maybe we should rewrite that part). Let me explain a little more:
- The logic here is to append extra tags using dummy images to non-x86_64 imagestreams. Normal images are already processed.
- If `cli` is embargoed, it will be missing in any public imagestreams (`x86_64`, `ppc64le`, `s390x`). When we attempt to omit a dummy for first time, `'cli' not in mirroring[key]` test (where key is one of `x86_64`, `ppc64le`, or `s390x`) is evaluated to `True`. In this case, all succeeding tests will always be `True`. `break` should have the same effects as `continue` but have better performance.
- For private imagestreams (`x86_64-priv`, `ppc64le-priv`, `s390x-priv`), if `cli` is missing, it will raise an error.

vfreex:
@sosiouxme I updated the dummy image substitution logic for better readability.

PR #244, opened 2020-07-17
[ART-1981] distinguish between enabled and shipping repos

joepvd:
There are new check that verify that if all the rpms on the images can
be accounted for, and that there are no unused repositories declared to
be enabled.

The repositories that are enabled, are configured through the
`enabled_repos` key in the image definition. Setting a repository here
has two effects:
1. It will be written in `.oit/{un,}signed.repo`, one of which will end
up as `/etc/yum.repos.d/ubi.repo` in the resulting image.
2. A metadata file in distgit, `content_sets.yml`, which is consumed by
TPS tests to verify if all the RPMs are accounted for.

This poses an issue with dependencies for BUILDER images, as the
repository definition file is for all images used during the build
process, and content_set.yml only affects the shipping artifact.

This change allows the following configuration in `images/*.yml`:
```yaml
enabled_repos:
- rhel-server-rpms
- rhel-server-optional-rpms
non_shipping_repos:
- rhel-server-optional-rpms
```

The end result is:
- both repositories are enabled in `/etc/yum.repos.d/`
- `rhel-server-optional-rpms` is _not_ present in `content_set.yml`

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -429,13 +429,16 @@ class ImageDistGitRepo(DistGitRepo):
 
         repos = self.runtime.repos
         enabled_repos = self.config.get('enabled_repos', [])
+        non_shipping_repos = self.config.get('non_shipping_repos', [])
+        shipping_repos = [repo for repo in enabled_repos if repo not in non_shipping_repos]
+
         for t in repos.repotypes:
             with self.dg_path.joinpath('.oit', f'{t}.repo').open('w', encoding="utf-8") as rc:
                 content = repos.repo_file(t, enabled_repos=enabled_repos)
                 rc.write(content)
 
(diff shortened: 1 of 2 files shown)


vfreex:
I would like to argue the assumption that content_set.yml only affects the shipping artifact.
At lease when building with ODCS, we need to put all repos required by the builder to content_set.yml, otherwise ODCS will not inject them to the builder container.

sosiouxme:
@vfreex you're right, but we might scrape by in 3.11 since we don't have many 2-stage builds.
However I think it's worth bringing this up to the CVP folks... those who are using ODCS will _have_ to include the content sets for both stages, and may have superfluous content sets for the shipping image.

sosiouxme:
I'll note that an alternate solution would be to build builder images separately so that they require no installs when used to build another image. But this solution is simpler.

PR #267, opened 2020-08-26
Raise an error if repo used in image meta is undefined

vfreex:
Some CVP content set errors are caused by misspelled repo names in image metas. However we didn't really validate them.

### DIFF
--- a/doozerlib/repos.py
+++ b/doozerlib/repos.py
@@ -271,7 +271,10 @@ class Repos(object):
         """Generates a valid content_sets.yml file based on the currently
         configured and enabled repos in the collection. Using the correct
         name for each arch."""
-
+        # check for missing repos
+        missing_repos = set(shipping_repos) - self._repos.keys()
+        if missing_repos:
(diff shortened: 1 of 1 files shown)


doozerlib/repos.py
```diff
@@ -271,7 +271,10 @@ def content_sets(self, shipping_repos=[]):
         """Generates a valid content_sets.yml file based on the currently
         configured and enabled repos in the collection. Using the correct
         name for each arch."""
-
+        # check for missing repos
+        missing_repos = set(shipping_repos) - self._repos.keys()
+        if missing_repos:
+            raise ValueError(f"Image meta references undefined repo(s): {missing_repos}")
```
joepvd, line 277: `s/Image meta/enabled_repos/`

joepvd (commented):
:thinking: Maybe the validator is a more appropriate place for this type of internal consistency check?

doozerlib/repos.py
vfreex: updated. next push

joepvd (approved):
/lgtm

:shipit:

PR #270, opened 2020-08-28
Fix non_shipping repos handling bug

vfreex:
If a repo is globally enabled for all images in group.yml,
non_shipping_repos will not have effects and the corresponding content sets will be added anyway.

This PR fixes that bug.


joepvd (approved):
/lgtm
Good catch!

PR #272, opened 2020-08-31
[ART-2186] bundles: Remove non-metadata files from manifests

thiagoalessio:
Apparently some repos contain additional files (YAML or not) that are
present in the manifests for other purposes, irrelevant to bundles, and
that is causing CVP tests to fail.


joepvd (commented):
Code looks good, but have little insight in if this is the adequate and necessary for the assignment. Copping out of approving.

PR #273, opened 2020-09-02
Bug 1867499: Pass KUBE_GIT_TREE_STATE=clean

joepvd:


doozerlib/metadata.py
```diff
@@ -395,6 +395,7 @@ def extract_kube_env_vars(self) -> Dict[str, str]:
                 godep_kube_minor = '0' if len(kube_version_fields) < 2 else kube_version_fields[1]
                 envs['KUBE_GIT_MINOR'] = f'{godep_kube_minor}+'  # For historical reasons, append a '+' since OCP patches its vendored kube.
                 envs['KUBE_GIT_COMMIT'] = kube_commit_hash
+                envs['KUBE_GIT_TREE_STATE'] = 'clean'
```
sosiouxme, line 398: bah. feels like we oughta be able to base this on something, but i see no obvious candidates. i'll wait for someone to file a bug telling us it said `clean` when it should have been `dirty` and they can tell us how they figured that out. (although i wouldn't imagine we would rebase on a dirty commit.)

joepvd: Agree. FWIW, I took inspiration for this approach from [here](https://github.com/openshift/doozer/blob/6bfec222d8b6606d78d495d7751467d11534450d/doozerlib/distgit.py#L1493)

PR #274, opened 2020-09-02
Migrate build datastore from simpledb to MariaDB/MySQL

adarshtri:
### Env Variables expected

1. DOOZER_DB_HOST
2. DOOZER_DB_PORT
3. DOOZER_DB_USER
4. DOOZER_DB_PASSWORD
5. DOOZER_DB_NAME

The names are self-explanatory. The variables are assumed to be present before doozer (dblib) is installed or used.

### DIFF
--- a/doozerlib/dblib.py
+++ b/doozerlib/dblib.py
@@ -1,15 +1,67 @@
 from __future__ import absolute_import, print_function, unicode_literals
+from future.utils import as_native_str
 import time
 import os
-import sys
 import threading
-import hashlib
-import boto3
 import traceback
+import mysql.connector as mysql_connector
+from doozerlib import constants
+import functools
 from .model import Missing
+import datetime
 
-# Records need unique names. Use this int atomically for extra protection.
-record_sequence = 0
+
+class DBLibException(Exception):
+
+    """
+    Exception class to record exceptions raised within the dblib module.
+    """
+
+    def __init__(self, message):
+        super(self.__class__, self).__init__(message)
+        self.message = message
+
+    @as_native_str
+    def __str__(self):
+        return self.message
+
+
+def try_connecting(func):
+    """
+    Decorator function, which when wrapped around tries to reconnect to mysql in case the connection
+    doesn't exist or is broken. Attempts 5 tries to reconnect before giving up.
+    Tries reconnecting in 25, 75, 150, 250, 375 seconds before giving up.
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+
+        db = args[0]
+
+        if db.mysql_db_env_var_setup:
+
+            if not db.connection or not db.connection.is_connected():
+                count = 1
+
+                while count <= 5:
+                    try:
+                        db.connection = mysql_connector.connect(
+                            host=os.getenv(constants.DB_HOST, constants.default_db_params[constants.DB_HOST]),
+                            user=os.getenv(constants.DB_USER),
+                            password=os.getenv(constants.DB_PWD),
+                            database=os.getenv(constants.DB_NAME, constants.default_db_params[constants.DB_NAME]))
+                        break
+                    except Exception as e:
+                        traceback.print_exc()
+                        db.runtime.logger.error("Unable to connect to mysql database.  Exception {}.".format(e))
+                        time.sleep(25 * count)
+                        count += 1
+
+            func(*args, **kwargs)
+        else:
+            db.connection = None
+
+    return wrapper
 
 
 class DB(object):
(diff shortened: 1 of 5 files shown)


sosiouxme (changes requested):
not a full review, just things i noticed while trying it out.

doozerlib/dblib.py
```diff
@@ -21,44 +148,128 @@ def __init__(self, runtime, environment, dry_run=False):
...
+
+        insert_string = f"insert into {table_name}({columns}) values({values})"
+
+        cursor = self.connection.cursor()
+        cursor.execute(insert_string)
+        self.connection.commit()
+        cursor.close()
+
+    @try_connecting
+    def create_payload_entry(self, payload, table_name):
```
sosiouxme, line 248: does this thread? we're definitely running builds in different threads, i would think we write db entries from threads as well...

sosiouxme: nevermind, i don't think dblib was originally designed to run threaded.

sosiouxme:
I ran this with an actual build (https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=31304314) and it completed, did not fail to connect etc (like it did before i fixed the NAME/PWD issue) but AFAICS did not create a record.

The ability to emit some kind of debugging info here would be helpful... I'm not sure what silently went wrong or just wasn't called.

doozerlib/dblib.py
```diff
@@ -1,17 +1,144 @@
...
+
+
+class IntegerIdentifier(AbstractTypeIdentifierHandler):
+    def identify(self, request: Any):
+        try:
+            int(request)
+
+            if len(request) <= 6:
+                return "INT", int(request)
+            return "BIGINT", int(request)
```
jupierce, line 62: Always use BIGINT. The first value I send you has no bearing on the size of the second value I send you.

doozerlib/dblib.py
```diff
@@ -21,44 +148,128 @@ def __init__(self, runtime, environment, dry_run=False):
...
+        if self.connection and self.connection.is_connected():
+            self.handle_missing_table(table_name)
+            self.handle_missing_columns(payload, table_name)
+            self.insert_payload(payload, table_name)
+            return 0
+        else:
+            return 1
+
+    @staticmethod
+    def rename_to_valid_column(column_name):
```
jupierce, line 258: I'd rather us just change the column names we use throughout the code. If I set an attribute that is invalid, throw an exception. That way, if I set an attribute, I know exactly what to include in my SQL without needing to remember this transformation.

doozerlib/dblib.py
```diff
@@ -21,44 +148,128 @@ def __init__(self, runtime, environment, dry_run=False):
...
+            WHERE table_name = '{0}'
+            """.format(table_name.replace('\'', '\'\'')))
+        if cursor.fetchone()[0] == 1:
+            cursor.close()
+            return True
+
+        cursor.close()
+        return False
+
+    def get_missing_columns(self, attr_payload, table_name):
```
jupierce, line 187: On the first pass through this method, cache the list of columns. Use the cache for subsequent calls to the method  and only add if a new column name does not appear in that cache. Add the new column to the cache on the way out of the method.

doozerlib/dblib.py
```diff
@@ -21,44 +148,128 @@ def __init__(self, runtime, environment, dry_run=False):
...
+    def create_column_identifier():
+        int_handler = IntegerIdentifier()
+        float_handler = FloatIdentifier()
+        datetime_handler = DateTimeIdentifier()
+        varchar_handler = VarcharIdentifier()
+
+        int_handler.set_next_handler(float_handler).set_next_handler(datetime_handler).set_next_handler(varchar_handler)
+        return int_handler
+
+    def check_table_exist(self, table_name):
```
jupierce, line 173: If table exists, remember that fact in a cache and don't query for it again.

doozerlib/dblib.py
```diff
@@ -21,44 +148,128 @@ def __init__(self, runtime, environment, dry_run=False):
...
+        self.connection.commit()
+        cursor.close()
+
+    @try_connecting
+    def create_payload_entry(self, payload, table_name):
+        if self.connection and self.connection.is_connected():
+            self.handle_missing_table(table_name)
+            self.handle_missing_columns(payload, table_name)
+            self.insert_payload(payload, table_name)
+            return 0
```
jupierce, line 253: Maybe?
```suggestion
            return self.insert_payload(payload, table_name)
```

doozerlib/dblib.py
```diff
@@ -21,44 +148,128 @@ def __init__(self, runtime, environment, dry_run=False):
...
+        for data in cursor:
+            existing_columns.add(data[0])
+
+        for column in attr_payload:
+            payload_columns.add(column)
+
+        missing_columns = list(payload_columns.difference(existing_columns))
+        return_columns = []
+        for column in missing_columns:
+            return_columns.append((column, self.column_identifier.identify(attr_payload[column])))
```
jupierce, line 203: As mentioned elsewhere, since mysql supports non-str data types, change Record.set so that it doesn't covert things to strings anymore. That should mean we don't need anything fancy to determine column type.
```python
def identify_column(val):
    pytype = type(val)
    if  pytype == 'int':
        return 'BIGINT'
    elif pytype == 'float':
        return 'FLOAT'
    elif pytype == `datetime.datetime`:
        return 'DATETIME'
    elif pytype == 'str':
        return 'VARCHAR'
    # .....etc
    else:
        raise IOError(f'Unexpect attribute type: {pytype}')       
```

Change all record set operations in the rest of the code to send in real ints, datetimes, etc.

jupierce (changes requested):
Nice work!

doozerlib/dblib.py
jupierce: Interacting with records should be threadsafe. https://github.com/openshift/doozer/blob/2bfe554eb350bd2219ea0258866489610911fb54/doozerlib/dblib.py#L165 provided the global lock for writing to simpledb. Ultimately, I think these methods should be `_create_payload_entry()` etc so that they are only invoked via Record.

doozerlib/runtime.py
```diff
@@ -241,6 +245,27 @@ def get_named_lock(self, absolute_path):
                 self.dir_locks[p] = new_lock
                 return new_lock
 
+    @staticmethod
+    def check_missing_db_env_var():
```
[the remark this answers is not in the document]
sosiouxme: That way essentially `--datastore <env>` informs the DB instead of `DB_NAME` ... I don't think we would want both. Can look up from a constant dict:
```
db_name_for_env = {
    "prod": "build",
    "int": "build_int",
}
```
That will ease the transition a bit too. I can go with that.

doozerlib/dblib.py
```diff
@@ -1,17 +1,144 @@
...
+                    count += 1
+
+        func(*args, **kwargs)
+
+    return wrapper
+
+
 class DB(object):
 
     def __init__(self, runtime, environment, dry_run=False):
```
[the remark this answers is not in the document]
adarshtri: What should I do here in case dry_run is set to true? Log the insert statement?

doozerlib/dblib.py
adarshtri: @jupierce I am  calling
 `db.create_payload_entry()` 
inside the
 `with self.db.runtime.get_named_semaphore('package::boto3'): `

Shouldn't that make all the inner function calls thread safe by default?

doozerlib/runtime.py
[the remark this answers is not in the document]
adarshtri: How do I access <env>? Is it available in runtime? Also assuming default databases or based on <env> should I create the database somewhere if is not exist or just assume it will exist? Otherwise the code starts getting exception everywhere in DBLib.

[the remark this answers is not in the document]
sosiouxme: pretty sure `--datastore` populates `runtime.datastore`

Something feels wrong about having doozer create the DB if not existing, but I'm not sure I can articulate it.

doozerlib/dblib.py
adarshtri: There is no concrete reasoning behind transforming the current column names to something. I am just trying to follow some suggested column naming conventions. So if we see the code I'm replacing . and - with _(underscore). Anything else for now passes the transformation. There is no need to remember anything. We can put any attribute name in the payload it is just replacing . and -.

doozerlib/dblib.py
```diff
@@ -161,28 +324,23 @@ def __enter__(self):
...
                 try:
-                    attr_payload = []
+                    attr_payload = {}
                     for k, v in self.attrs.items():
 
-                        if v is None or v is Missing:
+                        if v is None or v is Missing or v == '':
                             v = ''
+                        else:
+                            attr_payload[self.db.rename_to_valid_column(k)] = str(v)
```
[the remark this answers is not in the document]
adarshtri: Well I tried sticking to str assuming I don't know the sources from where the records will be coming. If record method is called from two places for one column (say column_x) and in first call the value provided in int and in second it is str then in that case going by the suggested logic (p_type = type(val) would give me conflicting type for column_x). That's why I chose to assign responsibility of type identification by dblib. And if the call with str is made before int but in reality the column is actually int and all other calls are coming with int we would be unnecessary storing column_x as varchar.

doozerlib/dblib.py
```diff
@@ -161,28 +324,23 @@ def __enter__(self):
         return self
```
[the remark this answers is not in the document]
adarshtri: I could see these fields of the payload but when I see simpledb there are lot more attributes for a build record whose source I didn't know. So just to avoid changing code at lot more places I tried doing transforming column names (to comply minimal convention) just before I am going to make an record entry into db table.

doozerlib/dblib.py
adarshtri: Agreed. But if there are two places where Record.set is called like this, Record.set("label.number", "1") first and Record.set("label.number", 2) after that and assuming label.number is a new column we would identify label.number as varchar instead of int for all the subsequent records.  There are possible flaws with my implementation as well but I guess we will have to handle some of the things manually as well since a lot of schema related things are happening dynamically which shouldn't happen ( like if the current implementation identifies a new column type as int but all the subsequent records are sending float for that value the implementation fails.)

jupierce: I consider a Record.set invocation with the wrong datatype a bug in doozer and nothing I think you should have to accommodate.

doozerlib/dblib.py
jupierce: It does. The suggested encapsulation just ensures that no one is tempted to call it directly in the future.

doozerlib/runtime.py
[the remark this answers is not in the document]
jupierce:
```
db_name_for_env = {
    "prod": "build",
    "int": "build_int",
}
```
No strong preference on the physical name for the database, but I suggested a generic `doozer_*` because I hope that the team will add more operation types over time. Right now there is just `build`, but `rebase`, `alignment_prs`, etc, can all provide valuable metrics.

doozerlib/dblib.py
sosiouxme: Having thought about this a little bit I have to agree with Justin. This kind of transform would only really make sense if we had a ton of legacy code that was going to need refactoring without it; we really don't. The only place there should be a column name transform is in migrating the data from simpledb to mysql... that aside, the column names used in the code should match exactly what ends up in the DB. If someone gives a bogus name (I had a look at the naming requirements, even with quoting some things are not allowed) it's probably best to detect that and fail fast (not wait for the db to give us an obtuse error).

doozerlib/dblib.py
```diff
@@ -1,64 +1,366 @@
...
+            if not dry_run:
+                cursor = self.connection.cursor()
+                cursor.execute(insert_string)
+                self.connection.commit()
+                cursor.close()
+                return 0
+            else:
+                print(insert_string)
+                return 0
+        except:
```
sosiouxme, line 329: despite the many existing bad examples, [we really should not have a bare except](https://moserei.de/2017/09/20/python-bare-except.html). I get really annoyed when I press `^C` (or hit "cancel" in jenkins) and it is ignored.
```suggestion
        except Exception:
```

doozerlib/dblib.py
```diff
@@ -161,28 +409,19 @@ def __enter__(self):
...
-                except:
-                    traceback.print_exc()
-                    self.runtime.logger.error(f'Unable to write to simpledb: {self.name}={self.attrs}')
+        with self.db.runtime.get_named_semaphore('dblib::mysql'):
+            try:
+                attr_payload = {}
+                for k, v in self.attrs.items():
+
+                    if v is None or v is Missing or v == '':
+                        v = ''
```
sosiouxme, line 418: this means the entry is entirely left out if the value is empty? if that's the intent... setting `v` at all is confusing. maybe:
```suggestion
                    if v is None or v is Missing or v == '':
                        continue
```

sosiouxme (changes requested):
Mostly stylistic suggestions. I was looking for why in my trial, out of 159 builds only 10 were recorded; it was not obvious to me.

Also, rebase again on upstream master... you have some dupe commits in there :)

doozerlib/dblib.py
```diff
@@ -1,64 +1,366 @@
...
+            return 1
+
+    @try_connecting
+    def create_payload_entry(self, payload, table_name, record_dry_run=False):
+
+        # this if condition only allows request to be written to db where dry_run is set
+        # false both at DB initiation in runtime and creating Record in the context
+        # manager. Any of the dry_run flags if set to true will not allow
+        # writing record to the database.
+        if not self.dry_run and not record_dry_run:
```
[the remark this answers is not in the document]
adarshtri: Won't this make a record entry if any of those two were True?  Did you mean
`if not (self.dry_run or record_dry_run):`

doozerlib/dblib.py
```diff
@@ -1,64 +1,366 @@
...
+
+        self._table_column_cache = {}
+
+    def check_missing_db_env_var(self):
+
+        """
+        The function checks if the expected environment variables are set. If not, raise an exception.
+        """
+
+        if not (constants.DB_USER in os.environ and constants.DB_PWD in os.environ and constants.DB_NAME in os.environ):
```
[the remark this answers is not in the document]
adarshtri: Noted. Yes, it should pick the default. I had deliberately avoided default for DB_NAME initially since I was not sure which MySQL will we test this on. And to be safe just make it must parameter. I'll switch it back to the default value.

doozerlib/dblib.py
```diff
@@ -1,64 +1,366 @@
...
+        if db.mysql_db_env_var_setup:
+
+            if not db.connection or not db.connection.is_connected():
+                count = 1
+
+                while count <= 5:
+                    try:
+                        db.connection = mysql_connector.connect(
+                            host=os.environ[constants.DB_HOST] if constants.DB_HOST in os.environ
+                            else constants.default_db_params[constants.DB_HOST],
```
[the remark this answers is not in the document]
adarshtri: Noted.

doozerlib/dblib.py
```diff
@@ -1,64 +1,366 @@
...
-            with self.lock:
-                if domain in self.existing_domains:
-                    # Already exists
-                    return
-                self.client.create_domain(
-                    DomainName=domain,
-                )
-                self.existing_domains[domain] = True
+        # mysql config for build records
+        self.mysql_db_env_var_setup = True
```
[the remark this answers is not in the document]
adarshtri: We don't. I'll remove them.

doozerlib/dblib.py
```diff
@@ -1,64 +1,366 @@
...
+
+        try:
+            columns, values = [], []
+
+            for column in payload:
+                columns.append(f'`{column}`')
+                values.append(f'"{payload[column]}"')
+
+            columns = ",".join(columns)
+            values = ",".join(values)
```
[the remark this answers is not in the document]
adarshtri: +1

doozerlib/dblib.py
adarshtri: @jupierce agreed this was a better approach. Incorporated this in the code. Now it's the responsibility of the client to send correct type during Record.set

doozerlib/dblib.py
adarshtri: Done. Yes makes sense.

doozerlib/dblib.py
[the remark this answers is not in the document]
adarshtri: Just when I was doing the changes I notices backquotes around column name. Any suggestions to achieve that with the short hand trick.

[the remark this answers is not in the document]
thegreyd:
```python
columns = ", ".join(f'`{column}`' for column in payload.keys())
```
can work

doozerlib/dblib.py
```diff
@@ -18,47 +70,330 @@ def __init__(self, runtime, environment, dry_run=False):
...
+
+        table_exist_status = self.check_table_exist(table_name)
+
+        if not table_exist_status:
+            self.runtime.logger.info("Table [{}] not found in database [{}]."
+                                     .format(table_name, self.db))
+            cursor = self.connection.cursor()
+            cursor.execute(f'create table {table_name}(log_{table_name}_id bigint auto_increment primary key)')
+            cursor.close()
+            self._table_column_cache[table_name] = dict()
```
thegreyd, line 272: Can do a set() here too

doozerlib/dblib.py
```diff
@@ -18,47 +70,330 @@ def __init__(self, runtime, environment, dry_run=False):
...
+        if missing_columns:
+            cursor = self.connection.cursor()
+
+            for column in missing_columns:
+                missing_column = column[0]
+                column_type = column[1][0]
+                cursor.execute(f"alter table {table_name} add column `{missing_column}` {column_type}")
+                self._table_column_cache[table_name][missing_column] = True
+
+                self.runtime.logger.info("Added new column [{}] of identified type [{}] to table [{}] "
```
thegreyd, line 297: It's interesting how we use a mix of f-strings and format() for formatting here, any particular reason? should we stick to one of them?

doozerlib/dblib.py
```diff
@@ -18,47 +70,330 @@ def __init__(self, runtime, environment, dry_run=False):
...
+
+            if not dry_run:
+                cursor = self.connection.cursor()
+                cursor.execute(insert_string)
+                self.connection.commit()
+                cursor.close()
+                return 0
+            else:
+                print(insert_string)
+                return 0
```
thegreyd, line 336: minor suggestion - we won't need the else part, plus keeping the shorter return branch in if 

```suggestion
            if dry_run:
                print(insert_string)
                return 0
            
            cursor = self.connection.cursor()
            cursor.execute(insert_string)
            self.connection.commit()
            cursor.close()
            return 0
```

doozerlib/dblib.py
```diff
@@ -18,47 +70,330 @@ def __init__(self, runtime, environment, dry_run=False):
...
+        This method is a wrapper method, which calls the method to check missing table, missing column and then
+        the method to insert record into table.
+        """
+
+        # this if condition only allows request to be written to db where dry_run is set
+        # false both at DB initiation in runtime and creating Record in the context
+        # manager. Any of the dry_run flags if set to true will not allow
+        # writing record to the database.
+
+        insert_status = None
```
thegreyd, line 359: Can we make this a boolean? (and in subsequent usage)

doozerlib/dblib.py
sosiouxme: We wrote a lot of code in python 2 when f-strings weren't available. Existing code and copy-pasta would likely get that.

Personally I greatly prefer f-strings.

sosiouxme:
@thegreyd i appreciate your comments here, would have liked to incorporate more but have kicked this down the road for too long. Merged and let's iterate further.

sosiouxme:
grrr... including that flake8 :-P

PR #277, opened 2020-09-08
release_calc_previous: Set minimum versions based on build-suggestions

sdodson:
We now have a mandate to ensure that z-stream upgrades are cumulative.
Overall it is also better for us to always offer and upgrade path that's
directly to the best version available. So rather than relying on a rolling
window of ten most recent z-streams and the two most recent previous minor
versions we'll use a static minimum versions and a block list sourced from OTA
managed content.

Prior to this change

$ ./doozer release:calc-previous --version 4.5.8
4.5.8,4.5.7,4.5.6,4.5.5,4.5.4,4.5.3,4.5.2,4.5.1,4.5.0-0.hotfix-2020-08-24-185832,4.4.20,4.4.19,4.4.18

After this change
$ ./doozer release:calc-previous --version 4.5.8
4.5.8,4.5.7,4.5.6,4.5.5,4.5.4,4.5.3,4.5.0-0.hotfix-2020-08-24-185832,4.4.20,4.4.19,4.4.18,4.4.17,4.4.16,4.4.15,4.4.14,4.4.13

Based on the following metadata
$ curl https://raw.githubusercontent.com/sdodson/cincinnati-graph-data/suggestions/build-suggestions/4.5.yaml
```
---
default:
  minor_floor: 4.4.13
  minor_block_list: []
  z_floor: 4.5.3
  z_block_list: []
s390x:
  minor_floor: 4.4.24
  minor_block_list: []
  z_floor: 4.5.3
  z_block_list: []
```

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -2102,13 +2102,13 @@ def release_calc_previous(version, arch, graph_url, graph_content_stable, graph_
 
     def get_build_suggestions(suggestions_url, major, minor, arch):
         """
-        Loads suggestions_url/major.minor.yaml and returns minor_floor, minor_block_list,
-        z_floor, and z_block_list
+        Loads suggestions_url/major.minor.yaml and returns minor_min, minor_max,
+        minor_block_list, z_min, z_max, and z_block_list
         :param suggestions_url: Base url to /{major}.{minor}.yaml
         :param major: Major version
         :param minor: Minor version
         :param arch: Architecture to lookup
-        :return: {minor_floor, minor_block_list, z_floor, z_block_list}
+        :return: {minor_min, minor_max, minor_block_list, z_min, z_max, z_block_list}
         """
         url = f'{suggestions_url}/{major}.{minor}.yaml'
         req = urllib.request.Request(url)
@@ -2124,12 +2124,13 @@ def release_calc_previous(version, arch, graph_url, graph_content_stable, graph_
     curr_versions, current_edges = get_channel_versions(candidate_channel)
     suggestions = get_build_suggestions(suggestions_url, major, minor, arch)
     for v in prev_versions:
-        if (semver.VersionInfo.parse(v) >= semver.VersionInfo.parse(suggestions['minor_floor'])
-                and semver.VersionInfo.parse(v) < semver.VersionInfo.parse(f'{major}.{minor}.0')
+        if (semver.VersionInfo.parse(v) >= semver.VersionInfo.parse(suggestions['minor_min'])
+                and semver.VersionInfo.parse(v) < semver.VersionInfo.parse(suggestions['minor_max'])
                 and v not in suggestions['minor_block_list']):
             upgrade_from.add(v)
     for v in curr_versions:
-        if (semver.VersionInfo.parse(v) >= semver.VersionInfo.parse(suggestions['z_floor'])
+        if (semver.VersionInfo.parse(v) >= semver.VersionInfo.parse(suggestions['z_min'])
+                and semver.VersionInfo.parse(v) < semver.VersionInfo.parse(suggestions['z_max'])
                 and v not in suggestions['z_block_list']):
             upgrade_from.add(v)
 
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -2039,7 +2039,10 @@ def config_gencsv(runtime, keys, as_type, output):
               help="Override content from stable channel - primarily for testing")
 @click.option("--graph-content-candidate", metavar='JSON_FILE', required=False,
               help="Override content from stable channel - primarily for testing")
-def release_calc_previous(version, arch, graph_url, graph_content_stable, graph_content_candidate):
+@click.option("--suggestions-url", metavar='SUGGESTIONS_URL', required=False,
+              default="https://raw.githubusercontent.com/openshift/cincinnati-graph-data/master/build-suggestions/",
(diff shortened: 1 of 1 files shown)


sdodson:
Depends on https://github.com/openshift/cincinnati-graph-data/pull/425

doozerlib/cli/__main__.py
```diff
@@ -2037,16 +2040,26 @@ def get_channel_versions(channel):
...
+    # z_floor, and z_block_list
+    url = f'{suggestions_url}/{major}.{minor}-{arch}.yaml'
+    req = urllib.request.Request(url)
+    req.add_header('Accept','application/yaml')
+    suggestions = yaml.safe_load(exectools.urlopen_assert(req))
     upgrade_from = set()
-    upgrade_from.update(get_channel_versions(prev_candidate_channel)[0][:2])
-    upgrade_from.update(get_channel_versions(prev_stable_channel)[0][:2])
+    prev_versions, prev_edges = get_channel_versions(prev_candidate_channel)
+    curr_versions, current_edges =get_channel_versions(candidate_channel)
```
wking, line 2051: nit: ` =get...` -> ` = get...` (missing space)

doozerlib/cli/__main__.py
```diff
@@ -2037,16 +2040,26 @@ def get_channel_versions(channel):
...
+    for v in prev_versions:
+        if ( semver.VersionInfo.parse(v) >= semver.VersionInfo.parse(suggestions['minor_floor']) and
+            semver.VersionInfo.parse(v) <  semver.VersionInfo.parse(f'{major}.{minor}.0') and
```
wking, line 2054: you probably don't need this filter, because you are looking at `prev_candidate_channel`.  But I'm in favor of keeping it, dropping `prev_candidate_channel` and parsing these out of `curr_versions` with this `minor`-cap.  Might have to do something to also exclude RCs and other pre-4.y.0 stuff...  Can we just look at the parsed minor value?

doozerlib/cli/__main__.py
```diff
@@ -2037,16 +2038,27 @@ def get_channel_versions(channel):
 
         return descending_versions, edges
 
+    # New rules, source an OTA managed datsource that specifies minor_floor, minor_block_list,
+    # z_floor, and z_block_list
+    url = f'{suggestions_url}/{major}.{minor}-{arch}.yaml'
+    req = urllib.request.Request(url)
```
wking, line 2044: Do we need a fallback for when this is not available?  Or do we treat that as "no floor" and include all versions?

sdodson: I'm not sure what the right thing to do in absence of data is. I need to test that they can still be overridden and if that's the case then I'd assume we just fall back to a human setting them and filing a ticket with us to get the data for a new release populated. For instance we should have populated 4.6 data by now.

wking: From /openshift/cincinnati-graph-data#425, as it stands today:

```yaml
z_floor: 4.6.0
```

Starting off there will block 4.5 -> `4.6.0-[rf]c.*` edges, won't it?  We want those to go into our pre-GA 4.6 releases.

sosiouxme:
The concept seems fine.

sdodson:
I'll take a pass at cleaning this up tomorrow but I think we need to hold this until we get the release controller to initiate tests for all upgrades in stable imageStream images.

sdodson:
/wip cancel
/hold
This is no longer WIP but we should hold this until the release controller runs all the requisite tests.

doozerlib/cli/__main__.py
sdodson: Updated that here https://github.com/openshift/cincinnati-graph-data/pull/425/files#diff-8703729926cb5013ad156f1586a9d1c8R7

sosiouxme:
```
$ doozer release:calc-previous --version 4.5.32
4.5.31,4.5.30,4.5.29,4.5.28,4.5.27,4.5.25,4.5.24,4.5.23,4.5.22,4.5.21,4.5.20,4.5.19,4.5.18,4.5.17,4.5.16,4.5.15,4.5.14,4.5.13,4.5.12,4.5.11,4.5.10,4.5.9,4.5.8,4.5.7,4.5.6,4.5.5,4.5.4,4.5.3,4.4.33,4.4.32,4.4.31,4.4.30,4.4.29,4.4.28,4.4.27,4.4.26,4.4.25,4.4.24,4.4.23,4.4.22,4.4.21,4.4.20,4.4.19,4.4.18,4.4.17,4.4.16,4.4.15,4.4.14,4.4.13
```
what stands out to me, aside from the length of the list, is that there are no hotfixes included... just wanted to double check if that seems right.

sdodson:
The hotfixes only get added as upgrades if they have less than two upgrade edges, here's the results I get from master.
```
$ ./doozer release:calc-previous --version 4.5.32
4.5.31,4.5.30,4.5.28,4.5.27,4.5.24,4.5.23,4.5.22,4.5.21,4.5.20,4.5.19,4.5.18,4.4.33,4.4.32,4.4.31
```

sosiouxme:
That explains why I see no hotfixes, but not all the extra old entries... investigating

sdodson:
Oh, that's the intent. 4.4.13 is the minimum for minor version upgrades, and 4.5.3 is the minimum for z-stream upgrades per the data maintained by OTA.
https://github.com/openshift/cincinnati-graph-data/blob/master/build-suggestions/4.5.yaml

sosiouxme:
yeah, i misunderstood... it's working fine. thanks :)

PR #278, opened 2020-09-08
Allow Dockerfile source to be a symlink

jupierce:
Allow upstream Dockerfiles to be symlinks. Previously, assertion of isfile threw an exception.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1821,8 +1821,11 @@ class ImageDistGitRepo(DistGitRepo):
         else:
             dockerfile_name = "Dockerfile"
 
-        # The path to the source Dockerfile we are reconciling against
-        source_dockerfile_path = os.path.join(self.source_path(), dockerfile_name)
+        # The path to the source Dockerfile we are reconciling against. Resolve to actual file
+        # if it is a symlink.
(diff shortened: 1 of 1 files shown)


doozerlib/distgit.py
```diff
@@ -1821,8 +1821,11 @@ def _merge_source(self):
         else:
             dockerfile_name = "Dockerfile"
 
-        # The path to the source Dockerfile we are reconciling against
-        source_dockerfile_path = os.path.join(self.source_path(), dockerfile_name)
+        # The path to the source Dockerfile we are reconciling against. Resolve to actual file
+        # if it is a symlink.
+        source_dockerfile_path = os.path.relpath(os.path.join(self.source_path(), dockerfile_name))
```
cgwalters, line 1826: Shouldn't this be `os.path.realpath()`?  `os.path.relpath` isn't documented as doing anything with symlinks.

cgwalters (changes requested):

doozerlib/distgit.py
cgwalters: Just `a` small difference in names there.

jupierce: Thanks, @cgwalters !

cgwalters (approved):
(I know nothing about this code and haven't looked at the surrounding context)
But looks plausible to me.

PR #282, opened 2020-09-12
Restrict the number of simultaneous distgit pushes

jupierce:
With every new release, we need to populate a huge amount of data into distgit the first time a build occurs. Dozens of `rhpkg push` invocations run simultaneously -- crushing distgit (or network i/o) and making each push take hours to complete. The simple 'timeout' used for these pushes, must be set absurdly high and is thus useless -- it currently set to 9800 seconds just so we could migrate to a new rhel-8 branch for 4.6.
This implementation limits pushes to 5 distgits at a time, which will allow us to set the push timeout to something sensible again.

Changes `get_named_lock` to `get_named_semaphore` to make this type of throttling easier elsewhere.

### DIFF
--- a/doozerlib/dblib.py
+++ b/doozerlib/dblib.py
@@ -162,7 +162,7 @@ class Record(object):
 
     def __exit__(self, *args):
         if not self.dry_run and self.db.client:
-            with self.db.runtime.get_named_lock('package::boto3'):
+            with self.db.runtime.get_named_semaphore('package::boto3'):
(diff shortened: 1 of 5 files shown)


sosiouxme (approved):
I'm wondering what effect this will have on "normal" pushes where we aren't populating a branch for the first time. But alright, let's see. It's likely not important.

sosiouxme:
Gotta fix the tests

PR #284, opened 2020-09-17
ART-2201: child images of embargoed images should be embargoed too

vfreex:
- When building on an already-existing base image, it looks at base image's release field. If it has a `.p1` suffix, the child image will be marked as embargoed as well.
- When building alongside the base image, it waits on the parent image's rebase thread before rebases itself. After the parent image is rebased, it reads parent image's rebase status and marks itself embargoed if the parent is embargoed.

It is not easy to test this PR. I tested it by:
- Choosing `ose-tools` and `openshift-enterprise-cli` for testing. `ose-tools`'s base image is `openshift-enterprise-cli`, and `openshift-enterprise-cli`'s base image is ``openshift-enterprise-base`. We will rebase `ose-tools` and `openshift-enterprise-cli` together and use the latest existing `openshift-enterprise-base` image as `openshift-enterprise-cli`'s base.
- Locating the line with `latest parent is embargoed` comment in the code change.
- Manually replacing `.p1` with `.p0` so it will think the latest existing `openshift-enterprise-base` image is embargoed.
- Rebasing `ose-tools,openshift-enterprise-cli` with
``` bash
$ ./doozer --debug --group=openshift-4.6 --latest-parent-version --images=ose-tools,openshift-enterprise-cli images:rebase --message=test --version=v4.6.0 --release=yuxzhu-test-1.p?
```
- You will see both of `ose-tools` and `openshift-enterprise-cli` have the `.p1` suffix.

## NOTE

When building a parent image and a child image together, we need the child image to wait for the parent image to determine if the parent has embargoed fixes. To simplify the implementation (without massive refactoring), this PR uses a similar locking approach like what we are doing in images:build build_container. But it seems this approach has a potential deadlock:
We are using Runtime.parallel_exec to start parallel tasks. It internally uses a thread pool with a capacity of the CPU core number. But, if all threads in the pool are waiting for parent image rebasing/building but no threads for parent images are started because no idle threads in the pool, it will dead lock.
But I think we haven't met the dead lock issue so far because we have more CPU cores than child images on buildvm. So the risk is probably low at the moment.

@sosiouxme @jupierce Could you guys have a look?

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -5,7 +5,7 @@ import shutil
 import time
 import traceback
 import errno
-from multiprocessing import Lock
+from multiprocessing import Lock, Event
 import yaml
 import logging
 import bashlex
@@ -299,6 +299,8 @@ class ImageDistGitRepo(DistGitRepo):
         super(ImageDistGitRepo, self).__init__(metadata, autoclone)
         self.build_lock = Lock()
         self.build_lock.acquire()
+        self.rebase_event = Event()
+        self.rebase_status = False
         self.logger = metadata.logger
         self.source_modifier_factory = source_modifier_factory
 
@@ -671,6 +673,20 @@ class ImageDistGitRepo(DistGitRepo):
         parent_dgr.wait_for_build(self.metadata.qualified_name)
         if terminate_event.is_set():
             raise KeyboardInterrupt()
+
+    def wait_for_rebase(self, image_name, terminate_event):
+        """ Wait for image_name to be rebased. """
+        image = self.runtime.resolve_image(image_name, False)
+        if image is None:
+            self.logger.info("Skipping image build since it is not included: %s" % image_name)
+            return
+        dgr = image.distgit_repo()
+        dgr.rebase_event.wait()
+        if not dgr.rebase_status:  # failed to rebase
+            raise IOError(f"Error building image: {self.metadata.qualified_name} ({image_name} was waiting)")
+        if terminate_event.is_set():
+            raise KeyboardInterrupt()
+        pass
 
     def build_container(
             self, profile, push_to_defaults, additional_registries, terminate_event,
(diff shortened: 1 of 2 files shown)


sosiouxme:
we [queue builds in tree-down order](https://github.com/openshift/doozer/blob/2531b5fcf172fff7a37f7f316f2d5a303c177cc1/doozerlib/cli/__main__.py#L1263-L1281) to ensure parents are queued before children (the child thread might still run at the same time but knows to wait) . we don't do the same for rebases, but i think we could; that should prevent deadlock, as long as we also account for parent rebase failure...

I think without that ordering, your concern about deadlocks is very valid.

thiagoalessio (approved):
/lgtm

I haven't tested locally, but the code looks good. If there is anything wrong, I'm sure we'll discover quickly on buildvm ;D

doozerlib/distgit.py
```diff
@@ -2000,43 +2021,54 @@ def extract_version_release_private_fix(self) -> (str, str, str):
...
-                self.private_fix = bool(prev_private_fix)  # preserve private_fix boolean for distgit-only repo
-
-            # Source or not, we should find a Dockerfile in the root at this point or something is wrong
-            assertion.isfile(df_path, "Unable to find Dockerfile in distgit root")
+    def rebase_dir(self, version, release, terminate_event):
+        try:
+            # If this image is FROM another group member, we need to wait on that group member to determine if there are embargoes in that group member.
+            image_from = Model(self.config.get('from', None))
+            if image_from.member is not Missing:
+                self.wait_for_rebase(image_from.member, terminate_event)
```
sosiouxme, line 2029: there is one member (`ose-metering-ansible-operator`) that uses other members as build stages (not just the final stage). i think this should loop through the `from` and the `builder` and wait on any members (and be private if any of them are private). i _think_ the ordering takes build stages into account. something like:

```suggestion
            parent_images = [Model(self.config.get('from', None))]
            parent_images.extend(self.config.get('from', {}).get('builder', [])))
            for image_from in parent_images:
                if image_from.member is not Missing:
                    self.wait_for_rebase(image_from.member, terminate_event)
```

Actually, that's not too pretty, and we'd have to do something similar at the part above determining if any of them are private... determining parents should be abstracted out.

sosiouxme (approved):
one minor edge case that would be nice to head off, but i think it's good to merge as it is.

doozerlib/distgit.py
vfreex: Yes, we are missing that part. I'll create a ticket in the backlog.

vfreex: https://issues.redhat.com/browse/ART-2349

PR #296, opened 2020-09-29
public_upstream_url is only populated when there is a difference between public & private

jupierce:


sosiouxme:
worked when i tested it

PR #298, opened 2020-10-01
Allow user to lock inputs to scan-sources

jupierce:
For scan-sources to be run repeatably and generate the same output, we need to be able to lock certain resources in time. This PR intends to implement 2, 3, and 4 below (`--upstream` already exists):
1. Upstream sources used:  Done with `--upstream`
2. Downstream distgit commits used: Done with `--downstream`
3. The brew/koji event: Done with `--brew-event`
4. Streams: Pullspec can be specified with `--stream`


doozerlib/runtime.py
```diff
@@ -335,6 +341,23 @@ def initialize(self, mode='images', clone_distgits=True,
         if no_group:
             return  # nothing past here should be run without a group
 
+        # For each "--stream alias image" on the command line, register its existence with
+        # the runtime.
+        for s in self.stream:
+            self.register_stream_override(s[0], s[1])
```
jupierce, line 347: Making `--stream` global for all commands.

doozerlib/runtime.py
```diff
@@ -1106,6 +1130,8 @@ def resolve_source(self, meta):
                         self.logger.info(f"Fetching commit-ish {meta.commitish} from upstream")
                         cmd = ["git", "-C", source_dir, "fetch", "origin", "--depth", "1", meta.commitish]
                         exectools.cmd_assert(cmd)
+                    exectools.cmd_assert(["git", "-C", source_dir, "checkout", meta.commitish])
```
jupierce, line 1133: @vfreex can you take a look at this? I was exercising `--upstream`, but found that it didn't leave the cloned source checked out at the expected commit.  Does this look appropriate to you?

PR #299, opened 2020-10-01
cmd_assert's on_retry receives string/list, not function

thiagoalessio:
https://github.com/openshift/doozer/blob/master/doozerlib/exectools.py#L82

This PR addresses the following error that eventually happens in the appregistry job:
```
AttributeError: 'function' object has no attribute 'read'
```
Example: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fappregistry/7143/console


joepvd (approved):
/lgtm

PR #300, opened 2020-10-01
Bump doozer

joepvd:
Prepare for release v1.2.25

2941e79 Add bump-doozer
fce62f3 cmd_assert's on_retry receives string/list, not function
fddb9ce Prepare to open PRs
62e80b0 public_upstream_url is only populated when there is a difference between public & private
dfee7f9 Add get_brew_buildinfo back This method is called by getattr(self, 'get_{}'.format(attr))()
f77224f Stdout is interpreted by pipeline; avoid it

And add script that produces :point_up: commit.


bump-doozer
```diff
@@ -0,0 +1,18 @@
...
+
+new_version=$( awk 'BEGIN{OFS=FS=".";}{$NF=$NF+1; print}' <<<"$old_version")
+
+echo "$new_version" >doozerlib/VERSION
+
+git add doozerlib/VERSION
+
+git commit -evm "Prepare for release v$new_version
+
+$(git log --oneline --no-merges v${old_version}..)"
```
sosiouxme, line 18: what about:
```suggestion
$(git log --oneline --no-merges v${old_version}..)"

git tag "v$new_version"
if [[ "$1" != "" ]]; then
    git push "$1" && git push "$1" "v$new_version"
done
```

sosiouxme (approved):

bump-doozer
joepvd: Tagging presumes that the proposal to cut a release is accepted and merged. Prefer to not touch that yet, and keep this a manual action :grimacing:

PR #301, opened 2020-10-05
[ART-2202] import machine-os-content into nightlies using build-sync

sosiouxme:
https://issues.redhat.com/browse/ART-2202

Enabling syncing m-os-c with gen-payload, and scanning for when an update is available.


requirements.txt
```diff
@@ -12,3 +12,4 @@ requests_kerberos
 semver
 boto3
 wrapt
+tenacity
```
thiagoalessio, line 15: :tada:

thiagoalessio (approved):
/lgtm

PR #309, opened 2020-10-21
[ART-2336] Don't rebuild bundles needlessly

thiagoalessio:


doozerlib/cli/__main__.py
```diff
@@ -2299,9 +2299,9 @@ def rebase_and_build_olm_bundle(runtime, operator_nvrs):
     def rebase_and_build(nvr):
         try:
             olm_bundle = OLMBundle(runtime)
-            olm_bundle.rebase(nvr)
+            did_rebase = olm_bundle.rebase(nvr)
             return {
-                'success': olm_bundle.build(),
+                'success': (olm_bundle.build() if did_rebase else True),
```
thiagoalessio, line 2304: just to confirm the evaluation order:
```
$ python
Python 3.8.5 (default, Aug 12 2020, 00:00:00)
[GCC 10.2.1 20200723 (Red Hat 10.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def foo():
...   print("foo func called")
...   return "foo"
...
>>> print( foo() if False else "bar" )
bar
>>> print( foo() if True else "bar" )
foo func called
foo
>>>
```

sosiouxme (approved):

PR #311, opened 2020-10-23
Enable functional tests and reduce scan-sources time by an order of magnitude

jupierce:
# Input Locking
Enables new doozer parameters which can lock its perspective in time for functional testing purposes:
- `--brew-event` make all calls to brew limited to the specified event (i.e. turn back the clock)
- `--group <group_name>[@commitish]` allows you to use ocp-build-data at a specific commit instead of HEAD
- `--lock-upstream <distgit> <commitish>` allows you to checkout a specific hash for the upstream of a given component
- `--lock-downstream <distgit> <commitish>` allows you to checkout a specific hash for the distgit of a given component
- `--stream <stream_name> <pullspec>` Allows you to override the normal resolution of a stream with a specific pullspec

# KojiWrapper
Each client session returned from Runtime (from methods like pooled_koji_client_session/shared_koji_client_session) will be a KojiWrapper instance. These instances intercept calls to the koji API enhance them in a few different ways.

## API Caching
`koji_api.getBuild(build_id, brew.KojiWrapperOpts(caching=True))` . The KojiWrapper will filter out any positional arguments that are of type KojiWrapperOpts. If caching=True, a global cache will be populated with the result of the API call if it returns successfully. If an identical call is invoked with caching=True, the cached result will be returned. This eliminates the need for custom caches and BlackBox LRU methods.

## API Logging
If a specific API is giving you trouble and you don't want to sort through the debug level logging of the koji client, you can use `koji_api.getBuild(build_id, brew.KojiWrapperOpts(logger=<logger>))`  to log the call and return results of the API invocation.

## Brew Event Locking Support
When executing in a --brew-event locked context, KojiWrapper api invocations will be automatically limited without the caller needing to be aware. If an API call is made which cannot be event # limited (i.e. the koji api does not accept an `event` or `beforeEvent` parameter, an exception will be raised. This can be bypassed by the caller explicitly telling the KojiWrapper that they understand the implication: `koji_api.getLastEvent(brew.KojiWrapperOpts(brew_event_aware=True))`. Use of this mechanism should be limited to when the caller knows for certain the result of the call will not impact the execution of doozer relative to the --brew-event specified. 

# Functional Tests
`tests_functional/` now contains a set of tests which exercise  doozer runs from an external process perspective. They leverage the input locking features to ensure predictable outputs from the runs. They are safe to run and have no external affects.
Execute with: `coverage3 run --branch --source doozerlib -m unittest discover -t . -s tests_functional`


thiagoalessio (approved):
/lgtm

sosiouxme:
I still haven't made time to understand this but I think it's lingered too long already and I'll trust Thiago's LGTM

PR #314, opened 2020-10-26
[ART-2336] Don't rebuild bundle if already exists for given NVR

thiagoalessio:
Tries to solve the following scenario:

> 1. foo-operator-4.6-42 builds
> 2. foo-operator-metadata-4.6.42-1 builds
> 3. foo-operator-4.6-43 builds
> 4. foo-operator-metadata-4.6.43-1 builds
> 5. We decide to ship foo-operator-4.6-42 and attach it to an advisory
> 6. Now we want to attach the corresponding bundles, so we run olm_bundles against the advisory... which does see changes when it rebases back to 4.6-42, and so builds and attaches foo-operator-metadata-4.6.42-2.

Example:
```
$ doozer -g openshift-4.6 olm-bundle:rebase-and-build cluster-logging-operator-container-v4.6.0-202010201329.p0 2>/dev/null
cluster-logging-operator-metadata-container-v4.6.0.202010201329.p0-2
```

### DIFF
--- a/doozerlib/olm/bundle.py
+++ b/doozerlib/olm/bundle.py
@@ -33,10 +33,7 @@ class OLMBundle(object):
         self.operator_nvr = operator_nvr
         self.get_operator_buildinfo()
 
-        builds = brew.get_tagged_builds(tags=[self.target],
-                                        build_type='image',
-                                        event=None,
-                                        session=self.brew_session)[0]
+        builds = self.brew_session.listTagged(tag=self.target, package=self.bundle_brew_component)
         vr = self.operator_nvr.replace(self.operator_brew_component, '')[1:].replace('-', '.', 1)
         bundle_nvrs = [build['nvr'] for build in builds]
         found = list(filter(lambda bundle_nvr: vr in bundle_nvr, bundle_nvrs))
--- a/doozerlib/olm/bundle.py
+++ b/doozerlib/olm/bundle.py
@@ -33,14 +33,15 @@ class OLMBundle(object):
         self.operator_nvr = operator_nvr
         self.get_operator_buildinfo()
 
-        _rc, out, _err = exectools.cmd_gather(
-            'brew list-builds --quiet --package={}'.format(self.bundle_brew_component)
-        )
-
+        builds = brew.get_tagged_builds(tags=[self.target],
+                                        build_type='image',
+                                        event=None,
+                                        session=self.brew_session)[0]
         vr = self.operator_nvr.replace(self.operator_brew_component, '')[1:].replace('-', '.', 1)
-        bundle_nvrs = [line.split(' ')[0] for line in out.split('\n')]
+        bundle_nvrs = [build['nvr'] for build in builds]
         found = list(filter(lambda bundle_nvr: vr in bundle_nvr, bundle_nvrs))
-        return found[-1].split(' ')[0] if found else None
+        found.sort(reverse=True)
+        return found[0] if found else None
 
     def rebase(self, operator_nvr):
         """Update bundle distgit contents with manifests from given operator NVR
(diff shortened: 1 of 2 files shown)


doozerlib/olm/bundle.py
```diff
@@ -24,6 +24,24 @@ class OLMBundle(object):
...
+        """Check if a bundle already exists for a given `operator_nvr`.
+
+        :param string operator_nvr: Operator NVR (format: my-operator-v4.2.30-202004200449)
+        :return string: NVR of latest found bundle build, or None.
+        """
+        self.operator_nvr = operator_nvr
+        self.get_operator_buildinfo()
+
+        _rc, out, _err = exectools.cmd_gather(
+            'brew list-builds --quiet --package={}'.format(self.bundle_brew_component)
```
sosiouxme, line 37: this will list every build since the beginning of time - i think we can safely restrict it to those which have been tagged for the release candidate branch? also we have `brew.get_tagged_builds()` for this which ought to be faster than shelling out.

sosiouxme (changes requested):

doozerlib/olm/bundle.py
thiagoalessio: much better!! done: https://github.com/openshift/doozer/pull/314/commits/50e050b7e9d6e7747e9a56ac25b4500772c7cb9b

doozerlib/olm/bundle.py
```diff
@@ -24,6 +24,25 @@ class OLMBundle(object):
...
+        :param string operator_nvr: Operator NVR (format: my-operator-v4.2.30-202004200449)
+        :return string: NVR of latest found bundle build, or None.
+        """
+        self.operator_nvr = operator_nvr
+        self.get_operator_buildinfo()
+
+        builds = brew.get_tagged_builds(tags=[self.target],
+                                        build_type='image',
+                                        event=None,
+                                        session=self.brew_session)[0]
```
sosiouxme, line 39: this gets all the images in the tag, which still seems excessive... and I forgot `brew.get_tagged_builds` really doesn't do what I expected here. This really should be simply
```suggestion
        builds = self.brew_session.listTagged(tag=self.target, package=self.bundle_brew_component)
```

doozerlib/olm/bundle.py
```diff
@@ -24,6 +24,25 @@ class OLMBundle(object):
...
+        vr = self.operator_nvr.replace(self.operator_brew_component, '')[1:].replace('-', '.', 1)
+        bundle_nvrs = [build['nvr'] for build in builds]
+        found = list(filter(lambda bundle_nvr: vr in bundle_nvr, bundle_nvrs))
```
sosiouxme, line 42: why work with strings when we have full records?
```suggestion
        found = list(build for build in builds if vr == build['version'])
```

doozerlib/olm/bundle.py
```diff
@@ -24,6 +24,25 @@ class OLMBundle(object):
...
+        found.sort(reverse=True)
+        return found[0] if found else None
```
sosiouxme, line 44: I really don't like string-sorting nvrs. it sets a bad example - numbers don't sort correctly as strings. even if that would actually work here (since it would be highly unlikely for us to see more than 10 bundle builds for any operator nvr).

but with a record-based approach this could be...
```suggestion
        found.sort(reverse=True, key=lambda build: int(build['release']))
        return found[0]['nvr'] if found else None
```

sosiouxme (changes requested):
sorry to mislead you :) 

i noticed some things along the way...

thiagoalessio:
Just tested again with all modifications applied, it works (much faster, BTW)
```
$ ./doozer -g openshift-4.6 olm-bundle:rebase-and-build cluster-logging-operator-container-v4.6.0-202010201329.p0 2>/dev/null
cluster-logging-operator-metadata-container-v4.6.0.202010201329.p0-2
```

sosiouxme:
:triumph:

PR #320, opened 2020-11-06
ART-1812: support building rpms against multiple targets

vfreex:
The idea here is getting rid of the `el8-rebuilds` job and handling
multiple target builds in Doozer.

With this PR, you can add the following options to ocp-build-data rpm
config:

```yaml
targets:
- rhaos-{MAJOR}.{MINOR}-rhel-7-candidate
- rhaos-{MAJOR}.{MINOR}-rhel-8-candidate
```

When running `doozer rpms:build`, Doozer will use tito to build an rpm
against the first target, then invoke Brew directly to build against
other targets using the same distgit commit.

Also added some unit tests (the existing unit test for RPMMetadata was
removed because it was broken and skipped for a long time).

### DIFF
--- a/doozerlib/brew.py
+++ b/doozerlib/brew.py
@@ -45,11 +45,11 @@ def get_watch_task_info_copy():
         return dict(watch_task_info)
 
 
-def watch_task(brew_hub, log_f, task_id, terminate_event):
+def watch_task(session, log_f, task_id, terminate_event):
     end = time.time() + 4 * 60 * 60
     watcher = koji_cli.lib.TaskWatcher(
         task_id,
-        koji.ClientSession(brew_hub, opts={'serverca': '/etc/pki/brew/legacy.crt'}),
+        session,
         quiet=True)
     error = None
     except_count = 0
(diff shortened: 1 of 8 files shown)


doozerlib/rpmcfg.py
```diff
@@ -357,36 +364,53 @@ def _build_rpm(self, scratch, record, terminate_event, dry_run=False):
...
+            secondary_targets = self.targets[1:]
+            if secondary_targets:
+                with self.runtime.shared_koji_client_session() as kcs:
+                    kcs.krb_login()
+                    source, _, _ = kcs.getTaskRequest(primary_task_id)
+                    for target in secondary_targets:
+                        self.logger.info(f"Submitting task for secondary target {target}")
+                        tid = kcs.build(source, target)
+                        task_ids.append(tid)
+                        task_url = "https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=" + str(tid)
```
sosiouxme, line 390: it's a minor thing, but here and below... i don't like seeing this embedded in code. it smells wrong.
for brewhub, we actually have a `group.yml` entry for the location, `self.runtime.group_config.urls.brewhub`
i don't remember why -- only real use (to us) for making it configurable would be to run a stage build.
i don't really like the idea of adding this to `group.yml` everywhere. i think it'd be nice to get `https://brewweb.engineering.redhat.com/brew/` out of the constants module (can make it configurable whenever we find an actual need for it).

doozerlib/rpmcfg.py
```diff
@@ -357,36 +364,53 @@ def _build_rpm(self, scratch, record, terminate_event, dry_run=False):
...
+                        self.logger.info("Build running: %s - %s - %s", self.rpm_name, target, task_url)
+            record["targets"] = self.targets
+            record["task_ids"] = task_ids
```
sosiouxme, line 393: hmm... did you actually try this with a db configured? i'm not sure simpledb accepts an array as a value, and i really don't expect mariadb does. perhaps they already auto-convert to strings if they see this but... may be wise to do it explicitly here?

i mean, ideally we'd get as many records as builds out of this, but i can see that would be tricky to wedge into this change without a significant refactor.

sosiouxme (commented):
mostly looks good, one thing i'd like to check you on.

this ended up seeming more complicated than i expected. thanks.

doozerlib/brew.py
```diff
@@ -81,6 +81,65 @@ def watch_task(brew_hub, log_f, task_id, terminate_event):
     return error
 
 
+def watch_tasks(brew_hub, log_f, task_ids, terminate_event):
+    if not task_ids:
+        return
+    end = time.time() + 4 * 60 * 60
+    session = koji.ClientSession(brew_hub, opts={'serverca': '/etc/pki/brew/legacy.crt'})
```
sosiouxme, line 88: this caught my eye. the whole reason for adding this oddly-specific CA was because we didn't understand how to handle `certifi` before. i don't think we need it any more and would rather not see it peppered around the code.

but more importantly, `watch_task` was the only method here that crafted its own client connection instead of accepting one as a parameter, and this is just a multiplexed copy of that. is there a good reason for wanting a bare client session? it seems to me that what's calling `watch_tasks(self.runtime.group_config.urls.brewhub, ...` would do better to call it with `self.runtime.build_retrying_koji_client()`.

doozerlib/rpmcfg.py
vfreex: Haven't tried this with DB. I'll take a look (or convert it to a json string if it takes time)

doozerlib/brew.py
vfreex: Right, I just copied from `watch_task` then modified on top of it. Better to use the `retrying_koji_client`

doozerlib/rpmcfg.py
vfreex: @sosiouxme I tested and revisited the simpledb code in Doozer. Currently Doozer only saves image build info into simpledb (see ImageDistGitRepo.update_build_db). This is only about rpms so it doesn't break.

vfreex:
@sosiouxme I've applied your review suggestions and rebased.

doozerlib/brew.py
```diff
@@ -45,11 +45,11 @@ def get_watch_task_info_copy():
         return dict(watch_task_info)
 
 
-def watch_task(brew_hub, log_f, task_id, terminate_event):
+def watch_task(session, log_f, task_id, terminate_event):
```
sosiouxme, line 48: I applaud this change but you'd need to update all the callers:
```
🐚 grep -rIe '\bwatch_task\b' doozer*
doozerlib/olm/bundle.py:        error = brew.watch_task(
doozerlib/operator_metadata.py:        return brew.watch_task(
doozerlib/distgit.py:from .brew import watch_task, get_build_objects
doozerlib/distgit.py:            error = watch_task(self.runtime.group_config.urls.brewhub, self.logger.info, task_id, terminate_event)
```

doozerlib/rpmcfg.py
sosiouxme: mkay. we'll probably want to start recording those before too long, but i suppose it's ok for now.

sosiouxme (changes requested):
otherwise LGTM

doozerlib/brew.py
vfreex: Oops, forgot to do that

vfreex:
PR updated to address the `watch_task` caller issue.

sosiouxme (approved):

PR #321, opened 2020-11-06
Update dev container configs

vfreex:
Pylance has replaced Microsoft Python Language Server as the recommended
Python language server for VSCode.


sosiouxme (approved):
Seems fine to me, but I don't use vscode so :shrug: any other vscode users want to try it?

PR #323, opened 2020-11-06
WIP: Always fetch the latest code and switch branches

vfreex:
So you don't need to run `doozer cleanup` before start a new build.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -139,8 +139,18 @@ class DistGitRepo(object):
 
             if os.path.isdir(self.distgit_dir):
                 self.logger.info("Distgit directory already exists; skipping clone: %s" % self.distgit_dir)
+                with Dir(self.distgit_dir):
+                    self.logger.debug(f"Checking out branch {distgit_branch}")
+                    # discard untracked files
+                    cmd = ["git", "-C", self.distgit_dir, "clean", "-fdx"]
+                    exectools.cmd_assert(cmd)
+                    # discard staged files
+                    cmd = ["git", "-C", self.distgit_dir, "checkout", "--force", "HEAD"]
+                    exectools.cmd_assert(cmd)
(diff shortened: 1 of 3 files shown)


sosiouxme (commented):
i feel slightly ambivalent about this one. it is sometimes useful to be able to futz with the existing contents of source or distgit repos when trying stuff out locally. in particular, `doozer --local images:rebase` used to be really useful like that (though it's been a long time since i used it and i doubt anyone else does). does that outweigh the surprise people get when they try to build the same thing again from altered/different groups and don't get updates? probably not...

joepvd (approved):
/lgtm

vfreex:
> i feel slightly ambivalent about this one. it is sometimes useful to be able to futz with the existing contents of source or distgit repos when trying stuff out locally. in particular, `doozer --local images:rebase` used to be really useful like that (though it's been a long time since i used it and i doubt anyone else does). does that outweigh the surprise people get when they try to build the same thing again from altered/different groups and don't get updates? probably not...

I never used `--local` option and am not sure if anyone actually builds against sources with local modifications. Right this would create confusion if anyone relies on that feature...

joepvd:
Would a global option `--(no-|)clean` help to decide between the two behaviors? Or is that too much option overhead for too little gain?
Or should we deprecate `images:rebase`, and start collecting breaking changes for 2.0?

jupierce:
Looks like we had some convergent thinking @vfreex . I saw this PR as I was opening: https://github.com/openshift/doozer/pull/337  :) . They seem very similar, but mine used an opt-in strategy (similar to what @joepvd came up with). Thoughts?

thiagoalessio (approved):
/lgtm

vfreex:
I will re-visit what is already covered by #337 and check if there's anything left in this PR can be integrated with #337 .

PR #324, opened 2020-11-09
ART-1812: Assert multi-target rpm builds have correct golang compilers

vfreex:
Includes https://github.com/openshift/doozer/pull/320

Use `check_golang_versions` option in ocp-build-data group.yml to
    specify how you would like to check them:
    
    - "x.y" (default): only major and minor version
    - "exact": the z-version must be the same
    - "no": do not check
    
    The default is `x.y` because our 4.6 and 4.7 targets currently have
    different z versions in their buildroots. We should change the check
    mode to `exact` once this is corrected.

### DIFF
--- a/doozerlib/rpmcfg.py
+++ b/doozerlib/rpmcfg.py
@@ -3,10 +3,13 @@ import io
 import os
 import re
 import shutil
+import threading
 import traceback
 from typing import Optional
 
-from doozerlib import util, constants
+import rpm
+
+from doozerlib import brew, constants, util
 from doozerlib.exceptions import DoozerFatalError
 from doozerlib.source_modifications import SourceModifierFactory
 
@@ -474,6 +477,14 @@ class RPMMetadata(Metadata):
                 # Status defaults to failure until explicitly set by succcess. This handles raised exceptions.
             }
 
+            if len(self.targets) > 1:  # for a multi target build, we need to ensure all buildroots have valid versions of golang compilers
+                self.logger.info("Checking whether this is a golang package...")
+                out, _ = exectools.cmd_assert(["rpmspec", "-q", "--buildrequires", "--", self.specfile])
+                if any(dep.startswith("golang") for dep in out.splitlines()):
+                    # assert buildroots contain the correct versions of golang
+                    self.logger.info("This is a golang package. Checking whether buildroots contain consistent versions of golang compilers...")
+                    self.assert_golang_versions()
+
             try:
                 def wait(n):
                     self.logger.info("Async error in rpm build thread [attempt #{}]: {}".format(n + 1, self.qualified_name))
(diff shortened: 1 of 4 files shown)


vfreex:
The rpm module should be installed by `rpm-py-installer`, and it should also be a dependency of `koji`. Not sure why it doesn't exist in CI. Needs to have a look.

vfreex:
Fixed CI by applying @joepvd 's patch https://github.com/openshift/elliott/pull/149 to Jenkinsfile.

doozerlib/rpmcfg.py
```diff
@@ -447,6 +473,13 @@ def build_rpm(self, version, release, terminate_event, scratch=False, retries=3,
                 # Status defaults to failure until explicitly set by succcess. This handles raised exceptions.
             }
 
+            if len(self.targets) > 1:  # for a multi target build, we need to ensure all buildroots have valid versions of golang compilers
+                # Check if this is a golang package
+                out, _ = exectools.cmd_assert(["rpmspec", "-q", "--buildrequires", "--", self.specfile])
+                if any(map(lambda dep: dep.startswith("golang"), out.splitlines())):
```
sosiouxme, line 479: this is fine but i think it's clearer as
```suggestion
                if any(dep.startswith("golang") for dep in out.splitlines()):
```

doozerlib/rpmcfg.py
```diff
@@ -479,6 +512,51 @@ def wait(n):
...
+    target_golang_versions_lock = threading.Lock()
+
+    def assert_golang_versions(self):
+        """ If this RPM component has multiple targets, check if all buildroots have consistent versions of golang compilers
+        """
+        check_mode = self.runtime.group_config.check_golang_versions or "x.y"  # no: do not check; x.y: only major and minor version; exact: the z-version must be the same
+        if check_mode == "no":
+            return
+        # target_golang_versions = {}  # a dict of target -> golang_version mappings
+        golang_components = ["golang", "golang-scl-shim"]
```
sosiouxme, line 525: i think you may want to add 
```suggestion
        golang_components = ["golang", "golang-scl-shim", "go-toolset"]
```

in el8 buildroots the RPM that actually gets installed and used is `go-toolset`. `golang` will probably be the same version but i'd rather check directly. i'm not certain this is the right thing to do though, maybe try it out and see. (we had a great testbed for golang mismatches in 4.4 and 4.5 until yesterday they were trued up...)

sosiouxme (commented):
suggestions, but otherwise LGTM

doozerlib/rpmcfg.py
vfreex: Do we rebuild `golang-scl-shim` for every `go-toolset` z version update? If no I think it is ok for golang-scl-shim to have differing z version...
For `go-toolset`, it seems to me the package name is `go-toolset-1.y` where y is `{12,13,14,15,...}` Don't really want to hardcode the y versions here. Seems I need to fetch the latest builds for all components with the `-build` tag then filter those prefixed with `go-toolset-*`...

sosiouxme: We don't rebuild `golang-scl-shim`, no. It might make sense to do that in the future for clarity, but I don't think we want to count on it yet. Every time we do that we have to ask someone to tag it in... that might also change though...

Filtering by prefix instead of using pkg name seems like a good strategy. Where we use inheritance to include go-toolset, we can have more than one minor version available; that's why golang-scl-shim hardcodes a specific version. E.g.:
```
$ brew latest-build rhaos-4.3-rhel-7-build go-toolset-1.1{0,1,2,3}
Build                                     Tag                   Built by
----------------------------------------  --------------------  ----------------
go-toolset-1.10-1.10.3-7.el7              devtools-2018.4-rhel-7  deparker
go-toolset-1.11-1.11.13-1.el7             devtools-2019.3-rhel-7  tstellar
go-toolset-1.12-1.12.12-4.el7             devtools-2019.4-rhel-7-candidate  deparker
```
For 4.4+ I have asked to just tag the specific builds in rather than inherit, but we haven't gone back to previous versions and done the same (probably won't unless there's some need for a change).

vfreex: Thanks for the clarification. I think the logic should be:
- get the `x.y` version of `golang-scl-shim`
- fetch the latest version of `go-toolset-${x}-${y}` where `x` and `y` are from `golang-scl-shim`.
- check whether the `x.y.z` version of `go-toolset-${x}-${y}` matches the one for RHEL8.
I'll update the PR to reflect the above logic.

vfreex: PR Updated

vfreex:
Rebased again.

sosiouxme (approved):
/lgtm
Let's merge all related PRs after release work is done tomorrow.

PR #335, opened 2020-12-07
ART-1080: update doozer:scan-sources to note config changes

vfreex:
For each image metadata, generate a digest that captures all relevant config. The current relevant items for images are:

- the metadata config itself (remove owners: though)
- the entries in group.yml attr repos: corresponding to the image's enabled_repos and non_shipping_repos attrs
- the entries in streams.yml corresponding to any stream: entries in the image from and from.builder attrs

When rebasing, create a dict with these entries, serialize it as json (with ordered keys so it's consistent), and store a digest of that in new distgit file .oit/config_digest.

When scan-sources runs, generate the digest the same way and compare with the distgit. If the distgit digest is different or missing, the image should be rebuilt.

### DIFF
--- a/doozerlib/cli/scan_sources.py
+++ b/doozerlib/cli/scan_sources.py
@@ -109,9 +109,10 @@ def config_scan_source_changes(runtime, ci_kubeconfig, as_yaml):
                     dgr = meta.distgit_repo()
                     path = Path(dgr.distgit_dir).joinpath(".oit", "config_digest")
                     if not path.exists():  # alway request a buiild if config_digest hasn't been stored
-                        runtime.logger.info('%s config_digest does not exist; request a build', dgk)
-                        needs_rebuild = True
-                        reason = '.oit/config_digest does not exist'
+                        pass  # TODO: uncomment the following after a week of natural rebuilds
+                        # runtime.logger.info('%s config_digest does not exist; request a build', dgk)
+                        # needs_rebuild = True
+                        # reason = '.oit/config_digest does not exist'
                     else:
                         with path.open('r') as f:
                             prev_digest = f.read()
--- a/doozerlib/cli/scan_sources.py
+++ b/doozerlib/cli/scan_sources.py
@@ -1,7 +1,9 @@
+from pathlib import Path
+
 import click
 import yaml
 
-from doozerlib import brew, rhcos, exectools
+from doozerlib import brew, exectools, rhcos
 from doozerlib.cli import cli, pass_runtime
 from doozerlib.cli import release_gen_payload as rgp
 from doozerlib.exceptions import DoozerFatalError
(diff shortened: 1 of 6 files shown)


doozerlib/image.py
```diff
@@ -657,6 +657,31 @@ def run_docker_cov(cmd):
...
+        builders = image_config.get("from", {}).get("builder", [])
+        from_stream = image_config.get("from", {}).get("stream")
+        referred_streams = {builder.get("stream") for builder in builders if builder.get("stream")}
+        if from_stream:
+            referred_streams.add(from_stream)
+        if referred_streams:
+            message["streams"] = {stream: streams[stream] for stream in referred_streams}
+
+        message = util.sort_dict(message)
+        digest = hashlib.sha256(json.dumps(message).encode("utf-8")).hexdigest()
```
sosiouxme, line 682: `sort_dict` is clever, but why not just do:
```suggestion
        digest = hashlib.sha256(json.dumps(message, sort_keys=True).encode("utf-8")).hexdigest()
```

doozerlib/cli/scan_sources.py
```diff
@@ -94,21 +103,28 @@ def add_assessment_reason(meta, changing, reason):
...
-                    add_assessment_reason(meta, needs_rebuild, reason=reason)
+                if not needs_rebuild:  # If no change has been detected, check configurations like image meta, repos, and streams to see if they have changed
+                    # We detect config changes by comparing their digest changes.
+                    # The config digest of the previous build is stored at .oit/config_digest on distgit repo.
+                    dgr = meta.distgit_repo()
+                    path = Path(dgr.distgit_dir).joinpath(".oit", "config_digest")
+                    if not path.exists():  # alway request a buiild if config_digest hasn't been stored
+                        runtime.logger.info('%s config_digest does not exist; request a build', dgk)
+                        needs_rebuild = True
+                        reason = '.oit/config_digest does not exist'
```
sosiouxme, line 114: I think your suggestion of ignoring missing `config_digest` to avoid the initial mass rebuild is good. Can we just:
```suggestion
                        pass  # TODO: uncomment the following after a week of natural rebuilds
                        # runtime.logger.info('%s config_digest does not exist; request a build', dgk)
                        # needs_rebuild = True
                        # reason = '.oit/config_digest does not exist'
```
... and plan to uncomment it next week?

sosiouxme (commented):
LGTM, some minor suggestions.

doozerlib/image.py
vfreex: That's convenient. Didn't realize there is a simpler way :)

vfreex:
Applied review suggestions and pushed a new commit.

sosiouxme (approved):
:shipit: after release work

vfreex:
I pushed another commit to add a check in case `owners` entry doesn't exist.

thiagoalessio (approved):
/lgtm

PR #337, opened 2020-12-14
Enable fast reclaimation of working_dir

jupierce:
Experiment to see if a new parameter --upcycle could be used to keep around and reuse 4.x specific working directories for jobs like scan-sources. This also avoids the long `rm -rf` runs associated with these workspaces.


vfreex:
@jupierce seems this has something related with https://github.com/openshift/doozer/pull/323/, which just cleans up the cloned repo, fetches , and resets to the latest remote branches. I could be wrong. Any thoughts on integrating them together?

PR #353, opened 2021-01-19
ART-2624: Fix scan-sources gets a different config digest than rebase

vfreex:
Some maintainer information is injected at runtime if not complete: https://github.com/openshift/doozer/blob/afc1bdb113935e30424ae06313524d2a974245a6/doozerlib/metadata.py#L330

By reading code calling get_maintainer_info, I think the maintainer
information is only retrieved from the return value and the changed
config meta is more like a side effect. So a simply fix could be copying
the dict before injecting into it.

### DIFF
--- a/doozerlib/metadata.py
+++ b/doozerlib/metadata.py
@@ -299,7 +299,7 @@ class Metadata(object):
         source_dir = self.runtime.resolve_source(self)
 
         # Maintainer info can be defined in metadata, so try there first.
-        maintainer = self.config.maintainer or dict()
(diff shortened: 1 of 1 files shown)


thiagoalessio:
CI failure has nothing to do with this PR, it was introduced with the MySQL stuff. Solved here: https://github.com/openshift/doozer/commit/f107a77afb004897de6fc8023fd798c10a7fb0c1

thiagoalessio (approved):
/lgtm

PR #359, opened 2021-02-25
Raise an error on PR creation failure

vfreex:
We see the following error on
https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/scheduled-builds/job/sync-ci-images/12692/console:

```
UnboundLocalError: local variable 'new_pr' referenced before assignment
```

Do not silently ignore PR creation exceptions. Let's see what will
happen if we re-raise that exception.


thiagoalessio (approved):
/lgtm

jupierce:
Great catch! Just a note on exceptions.. if you are re-raising an exception, just use `raise` -- otherwise it will change with your stack trace. Adding that change into https://github.com/openshift/doozer/pull/358/

PR #362, opened 2021-03-03
Pin target release for all operator builds:

jupierce:
https://docs.engineering.redhat.com/display/CFC/Delivery describes
that 'v4.6' is actually interpreted as 'v4.6+'. To pin to an
exact release, we must use `=<version>`.


sosiouxme (approved):
suits!

PR #365, opened 2021-03-18
[ART-2727] add select function

Ximinhan:
Currently in mysql database table log_build in dev_build db, direct doozer get the same result like it used to query

test job: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/ximhan-check/job/master/39/console
slack link: https://coreos.slack.com/archives/G01HDTRUU3T/p1618225276004600

### DIFF
--- a/doozerlib/dblib.py
+++ b/doozerlib/dblib.py
@@ -123,6 +123,18 @@ class DB(object):
         self.runtime.logger.info("Found all environment variables required for db setup.")
         return True
 
+    def select(self, expr, limit):
+        exeresult=""
+        if self.mysql_db_env_var_setup:
+            db_connection = mysql_connector.connect(host=self.host,
+                                                      user=self.db_user,
+                                                      password=self.pwd)
+            cursor = db_connection.cursor()
+            cursor.execute("{} LIMIT {}".format(expr,limit))
+            exeresult = cursor.fetchall()
+            cursor.close()
+        return exeresult
+
     def check_database_exists(self):
 
         """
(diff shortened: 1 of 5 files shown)


thiagoalessio (approved):
/lgtm

doozerlib/dblib.py
```diff
@@ -123,6 +123,24 @@ def check_missing_db_env_var(self):
         self.runtime.logger.info("Found all environment variables required for db setup.")
         return True
 
+    def select(self, expr, limit):
```
sosiouxme, line 126: could this have a comment? just looking at this naively, i'd want to know:

1. why do we need this at all
2. what's the format of `expr`
3. why not have a default `limit=100` or something

sosiouxme (changes requested):
can we have unit or functional tests that exercise this?

adarshtri:
I think there will changes in doozerlib/cli/__main__.py without which select most probably wouldn't work. Doozer cli is the entry point for image health job to get data.

doozerlib/dblib.py
```diff
@@ -123,6 +123,36 @@ def check_missing_db_env_var(self):
         self.runtime.logger.info("Found all environment variables required for db setup.")
         return True
 
+    def select(self, expr, limit=100):
```
adarshtri, line 126: The "expr" here will get right column names to be projected, right where condition and correct order by clause as well. But it will have wrong table name. [This](https://github.com/openshift/doozer/blob/3651eba338c1792f666945cfedb824a20e587d03/doozerlib/cli/__main__.py#L231) method will mostly require changes. It is expecting older configuration which might not be present with new implementation. Also it is referring to wrong table to fetch data from. The current table name is (log_build).

adarshtri (changes requested):

PR #366, opened 2021-03-19
Allow builder/base images to be full pullspecs

jupierce:
Previously, scan-sources would try to evaluate a full pullspec as a org/repo relative to brew and an exception would be thrown in a case like: https://github.com/openshift/ocp-build-data/commit/fd33bef0935dd0bff6dc06154d9688832d6a912d#diff-5d1723b2670c56bbd766f379858f004b5b0067fd6c8bfccc7271151e4b3ddd98R16


vfreex (approved):
/lgtm

PR #406, opened 2021-05-15
expand arch translation to include aarch64<->arm64

sosiouxme:
now that ARM is imminent we need doozer to understand its arch names. turns out it's just as special as x86_64/amd64 that way. this puts a central translation framework in place.


vfreex (commented):
/lgtm

vfreex (approved):

PR #407, opened 2021-05-17
[ART-2925] Pin brew repo-urls to distgit commit

vfreex:
- When rebasing distgit, waits for cgit cache to include the distgit
  commit. Raises an error if the cache does not update after 10 minutes.
- Before building an image, queries cgit for the existence of the .oit repo file for the commit. Raise an error if it doesn't exist.
- When building an image, supplies an .oit repo repo-url to brew that is
  pinned to the distgit commit that brew task is building.


jupierce (approved):
Looks great! One suggestion to consider.

vfreex:
Thanks. Review suggestion applied.

thiagoalessio (approved):
/lgtm

PR #408, opened 2021-05-17
Initial implementation of --assembly

jupierce:
First part of https://issues.redhat.com/browse/ART-2940 .
Basic implementation of doozer `--assembly` parameter. It does not impact builds until group.yml is updated with
```
assemblies:
  enabled: true
```

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1430,7 +1430,7 @@ class ImageDistGitRepo(DistGitRepo):
                                 self.logger.info('[{}] parent image {} not included. Looking up FROM tag.'.format(self.config.name, base))
                                 base_meta = self.runtime.late_resolve_image(base)
                                 _, v, r = base_meta.get_latest_build_info()
-                                if r.endswith(".p1"):  # latest parent is embargoed
+                                if util.isolate_pflag_in_release(r) == 'p1':  # latest parent is embargoed
                                     self.private_fix = True  # this image should also be embargoed
                                 mapped_images.append("{}:{}-{}".format(base_meta.config.name, v, r))
                             # Otherwise, the user is not expecting the FROM field to be updated in this Dockerfile.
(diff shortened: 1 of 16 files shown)


doozerlib/runtime.py
```diff
@@ -399,6 +400,14 @@ def initialize(self, mode='images', clone_distgits=True,
         self.group_dir = self.gitdata.data_dir
         self.group_config = self.get_group_config()
 
+        if self.group_config.assembles.enabled or self.enable_assemblies:
+            if re.match(r'^[a-zA-Z0-9._]+$', self.assembly) is None:
+                raise ValueError('Assembly names may only consistent of alphanumerics, ., and _. ')
```
vfreex, line 405: Would also like to disallow `self.assembly` to start or end with `.`:
```suggestion
            if re.fullmatch(r'[\w.]+', self.assembly) is None or self.assembly[0] == '.' or self.assembly[-1] == '.':
                raise ValueError('Assembly names may only consistent of alphanumerics, ., and _, but not start or end with a dot (.).')
```

doozerlib/distgit.py
```diff
@@ -1514,8 +1517,14 @@ def update_distgit_dir(self, version, release, prev_release=None):
                     release = release[:-3]  # strip .p?
                     release += pval
 
+                if self.runtime.assembly:
+                    release += f'.assembly.{self.runtime.assembly}'
```
vfreex, line 1521: I realized this breaks the assumption that `.p[0|1]` is always at the end of the `release` field.
We need to find & replace all `endswith(".p1")` `endswith(".p0")` calls in Doozer code with `find()` or `in` statement.

vfreex (changes requested):
With assemblies enabled, release field will not end with `.p0` or `.p1`. We need to replace all code that explicitly checks `.p0|1` value using the `string.endswith` call with substring find. Otherwise LGTM.

doozerlib/runtime.py
jupierce: Good call. Updated.

doozerlib/distgit.py
jupierce: I think I caught them all - everything should be using `util.isolate_pflag_in_release` now.

sosiouxme (approved):
just make flake8 happy

vfreex (approved):
This is great! Thanks for the update.

/lgtm

vfreex:
Merging this (Luke has approved us taking some risk and merging)

PR #409, opened 2021-05-19
Fix a typo that prevents assemblies from working

vfreex:


vfreex:
I am going to merge this because it is straightforward.

PR #411, opened 2021-05-21
ART-2941: Basic interpretation of assembly in gen-payload

vfreex:
- If assemblies are disabled, everything in `-candidate` tags will be
introduced into nightlies (existing behavior).

- If assemblies are enabled, only images containing `.assembly.<assembly-name>` in its `RELEASE` field will be
introduced into nightlies. If no build for an image is found, the newest
build containing `.assembly.stream` will substitute.

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -214,6 +214,9 @@ class PayloadGenerator:
                 if not chosen_build and self.runtime.assembly != "stream":
                     # If no such build, fall back to the newest build containing ".assembly.stream"
                     chosen_build = next((build for build in builds if build["release"].endswith(".assembly.stream")), None)
+                if not chosen_build:
+                    # If none of the builds have .assembly.stream in the RELEASE field, fall back to the latest build without .assembly in the RELEASE field
+                    chosen_build = next((build for build in builds if ".assembly." not in build["release"]), None)
             brew_latest_builds.append(chosen_build)
 
         # look up the archives for each image (to get the RPMs that went into them)
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -200,7 +200,7 @@ class PayloadGenerator:
         :return: list of build records, list of images missing builds
         """
         tag_component_tuples = [(image.candidate_brew_tag(), image.get_component_name()) for image in payload_images]
-        lists_of_brew_builds = brew.get_tagged_builds(tag_component_tuples, "image", self.brew_event, self.brew_session)
+        lists_of_brew_builds = brew.get_tagged_builds(tag_component_tuples, "image", event=self.brew_event, session=self.brew_session, inherit=True)
         brew_latest_builds = []
         for builds in lists_of_brew_builds:  # builds are ordered from newest tagged to oldest tagged
             chosen_build = None
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -1,13 +1,17 @@
 import io
 import json
+from typing import List, Optional, Tuple
 
 import click
 import yaml
+from koji import ClientSession
 
 from doozerlib import brew, build_status_detector, exectools, rhcos, state
 from doozerlib.cli import cli, pass_runtime
 from doozerlib.exceptions import DoozerFatalError
-from doozerlib.util import red_print, yellow_print, go_suffix_for_arch
+from doozerlib.image import ImageMetadata
+from doozerlib.runtime import Runtime
+from doozerlib.util import go_suffix_for_arch, red_print, yellow_print
 
 
 @cli.command("release:gen-payload", short_help="Generate input files for release mirroring")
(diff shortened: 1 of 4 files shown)


doozerlib/cli/release_gen_payload.py
```diff
@@ -170,16 +189,32 @@ def _get_payload_images(self, images):
...
+        If assemblies are enabled, it will return the latest tagged brew build in the given assembly. If no such build, it will fall back to the latest build in "stream" assembly.
+
         :param payload_images: a list of image metadata for payload images
         :return: list of build records, list of images missing builds
         """
         tag_component_tuples = [(image.candidate_brew_tag(), image.get_component_name()) for image in payload_images]
-        brew_latest_builds = brew.get_latest_builds(tag_component_tuples, "image", self.brew_event, self.brew_session)
-        # there's zero or one "latest" build in each list; flatten the data structure.
-        brew_latest_builds = [builds[0] if builds else {} for builds in brew_latest_builds]
+        lists_of_brew_builds = brew.get_tagged_builds(tag_component_tuples, "image", self.brew_event, self.brew_session)
```
jupierce, line 203: I found this API that allows specifying globs to match NVRs which may help cut down on the data transfer:
https://koji.fedoraproject.org/koji/api

```
listBuilds(packageID=None, userID=None, taskID=None, prefix=None, state=None, volumeID=None, source=None, createdBefore=None, createdAfter=None, completeBefore=None, completeAfter=None, type=None, typeInfo=None, queryOpts=None, pattern=None)
  Return a list of builds that match the given parameters

        Filter parameters
            - packageID: only builds of the specified package (numeric id)
            - userID: only builds owned by the given user (numeric id)
            - taskID: only builds with the given task ID
                If taskID is -1, only builds with a non-null task id
            - volumeID: only builds stored on the given volume (numeric id)
            - source: only builds where the source field matches (glob pattern)
            - prefix: only builds whose package name starts with that prefix
            - pattern: only builds whose nvr matches the glob pattern
            - state: only builds in the given state (numeric value)
```

Using a glob in `pattern` we might more efficiently query. 

1. Search for `*.assembly.{name}` 
2. If no results, search for, `*.assembly.stream`  

Getting all tagged builds for all images could return tens-of-thousands of records for an aging release. Thoughts?

I enabled brew-event pinning for the API here: https://github.com/openshift/doozer/pull/413.

doozerlib/cli/release_gen_payload.py
```diff
@@ -170,16 +189,32 @@ def _get_payload_images(self, images):
...
         return payload_images, invalid_name_items
 
-    def _get_latest_builds(self, payload_images):
+    def _get_latest_builds(self, payload_images: List[ImageMetadata]) -> Tuple[List[BuildRecord], List[ImageMetadata]]:
         """
-        find latest brew build (at event, if given) of each payload image.
+        Find latest brew build (at event, if given) of each payload image.
+
+        If assemblies are disabled, it will return the latest tagged brew build with the candidate brew tag.
+        If assemblies are enabled, it will return the latest tagged brew build in the given assembly. If no such build, it will fall back to the latest build in "stream" assembly.
```
jupierce, line 197: If `assembly.stream` does not exist, we should fall back to `builds[0]`. Right now, vertical pod autoscaler is failing to build (https://github.com/openshift/vertical-pod-autoscaler-operator/pull/91). Seems reasonable to fall back to the actual latest build as a last resort.

sosiouxme (commented):
nothing to add to justin's suggestions. seems straightforward enough.

doozerlib/cli/release_gen_payload.py
vfreex: @jupierce `builds[0]` could be in a different assembly which is not `assembly.<runtime.assembly>` or `assembly.stream`. I think we'd better to default to the latest build without `.assembly.` in its release?

doozerlib/cli/release_gen_payload.py
vfreex: I am trying the `listBuilds` and also comparing the numbers. Will put my findings later.

vfreex:
@jupierce I tried the brew `listBuilds` API and compared it with `listTagged`. Here are some findings:

1. `listBuilds` doesn't have a parameter to filter by tag. So the API actually returns pattern matched builds for all tags. We have to call `listTags` for each build to determine whether it is for the current release. Currently `len(brew_session.listBuilds(pattern="*.assembly.stream"))` is `1257`. I think we can assume in the future most of our builds are for the `stream` assembly and custom (patch) assembly builds are less, so without a tag filter `listBuilds` could return much more results than `listTagged`.
2. For each component, if there's no builds having `.assembly.<name>` in its release field, we need to fall back to `.assembly.stream`, then the latest build without `.assembly`. The fallback process needs separate API calls. But for `listTags`, it is just 1 call, which could reduce the wait time.
3. Following are the numbers of builds returned by `listTagged(tag, build_type="image", inherit=False)` for each OCP minor version:
**rhaos-4.8-rhel-8-candidate + rhaos-4.8-rhel-7-candidate**: 4750
**rhaos-4.7-rhel-8-candidate + rhaos-4.7-rhel-7-candidate**: 1342
**rhaos-4.6-rhel-8-candidate + rhaos-4.6-rhel-7-candidate**: 1358
**rhaos-4.5-rhel-8-candidate + rhaos-4.5-rhel-7-candidate**: 1070
**rhaos-3.11-rhel-8-candidate + rhaos-3.11-rhel-7-candidate**: 954

I guess 3.11-4.7 have less candidate builds than 4.8 because they have less frequent code changes and old builds have been untagged by GC.

So I would like to keep using the `listTagged` API until we hit some bottlenecks. What do you think?

vfreex:
I updated the PR to fall back to the latest builds with no `.assembly.` in its release field as the last resort. Let me know if this is the correct behavior.

doozerlib/cli/release_gen_payload.py
jupierce: That's a good idea. Should never happen in our flows, but makes sense for completeness.

jupierce:
@vfreex - I had not noticed the api didn't accept tag. Thanks for the analysis. Certainly no need to worry about optimization at this point.
For future consideration, `queryOpts` takes a `limit` and `orderBy` key to control the items returned. I think a glob of `{component-name}-v{major}.{minor}.*.assembly.{name}' with limit=1 should get us the exact component build we are looking (I can't think of a reason we would need to check the tags).

jupierce (approved):

vfreex:
@jupierce I didn't notice the usage of `queryOpts`. Yeah, it should be quick to to use a glob like that with `queryOpts`. 
Checking tags are just to ensure the build has the -candidate tag. I was not sure if it is ok to completely ignore the tags (since we used to untag things from -candate to exclude them from nightlies and hotfix builds may not have the -candidate tag). I will leave a comment on the JIRA for the potential path to future optimization. Thank you for hint.

PR #412, opened 2021-05-21
update setup.py to exclude test_functional package

vfreex:


vfreex:
merging because this is harmless.

PR #414, opened 2021-05-25
[ART-2626] support image list used in detect-embargo

Ximinhan:
`doozer detect-embargo rpms` command and `doozer detect-embargo images` command support  check whether latest build of given rpm/image name become embargo.


vfreex (changes requested):
Currently I don't have a full picture of the issue so I am not sure if a new verb is actually needed. I will take a look at that later this week.

But there's an issue in converting distgit names to brew component names.

doozerlib/cli/detect_embargo.py
```diff
@@ -45,6 +45,29 @@ def detect_in_nvr(runtime: Runtime, nvrs, as_yaml, as_json):
     embargoed_builds = detect_embargoes_in_nvrs(runtime, nvrs)
     print_result_and_exit(embargoed_builds, None, None, as_yaml, as_json)
 
+@detect_embargo.command("images", short_help="Detect embargoed fixes in given images")
```
vfreex, line 48: The verb name `detect-embargo images` sounds like it is to detect embargoes in a list of images, but what it actually does is finding images that changed to embargoes from non-embargoes.

doozerlib/cli/detect_embargo.py
```diff
@@ -141,6 +164,31 @@ def detect_embargoes_in_nvrs(runtime: Runtime, nvrs: List[str]):
...
+    :param images: list of images name
+    :return: list of Brew build dicts that have embargoed fixes
+    """
+    runtime.logger.info(f"Fetching {len(images)} builds from Brew...")
+    brew_session = runtime.build_retrying_koji_client()
+    detector = bs_detector.BuildStatusDetector(brew_session, runtime.logger)
+    embargoed_builds = []
+    for image in images:
+        pattern = image + '-container-v' + version
+        results = brew_session.search(pattern, 'build', 'regexp', {'limit': 2, 'order': '-id'})
```
vfreex, line 179: Hmm... I think ordering by id might not be 100% correct. Maybe we should order by the time when a build is tagged in.

vfreex (changes requested):

doozerlib/cli/detect_embargo.py
Ximinhan: I tried to use time but seems brew don't support such query options

vfreex:
This PR is almost there but there are still some issues:
1. `brew_session.search` used here to search for build is not compatible with assemblies (mainly because it is not aware of the `.assembly.$name` suffix in build NVRs). I think we should only care about the `stream` assembly in this PR. @ximhan could you have a look at `Metadata.get_latest_build` for how to get the latest 2 builds for an assembly? You can either write a new function to return the recent 2 builds using a similar logic like `Metadata.get_latest_build` or directly modify that method to return recent 2 builds. Let me know if you have any questions.
2. The verb name `detect-embargo images` doesn't sound like a good name. It sounds like the verb is to detect embargoes in images but actually it is to detect changes. Would like to a better name instead.

PR #417, opened 2021-05-27
Include git hash in image release

jupierce:
Doing so will help us prevent having to clone distgits
during scan-sources.


sosiouxme (approved):
YOLO

PR #420, opened 2021-05-29
Find current metas by uuid tag instead of nvr

jupierce:
The previous logic assumed that the other images used
the same version & release field as the operator
being built. However, releases now contain .git.<upstream hash>
in their release. Use the uuid tag placed in additional
tags, which should serve the same purposes but more reliably.


jupierce:
Operator bundle builds are broken due to https://github.com/openshift/doozer/pull/417 . Continuing the yolo.

PR #423, opened 2021-06-02
[ART-2974] Prevent unnecessary clone to determine RPM package name

jupierce:
https://issues.redhat.com/browse/ART-2974

### DIFF
--- a/doozerlib/rpmcfg.py
+++ b/doozerlib/rpmcfg.py
@@ -207,10 +207,11 @@ class RPMMetadata(Metadata):
                 if (check_mode == "exact" and nevr[2] != first_nevr[2]) or (check_mode == "x.y" and nevr[2].split(".")[:2] != first_nevr[2].split(".")[:2]):
                     raise DoozerFatalError(f"Buildroot for target {target} has inconsistent golang compiler version {nevr[2]} while target {first_target} has {first_nevr[2]}.")
 
-    def get_package_name(self, default=-1):
+    def get_package_name_from_spec(self, default=-1):
         """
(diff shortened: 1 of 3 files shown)


doozerlib/rpmcfg.py
```diff
@@ -233,6 +234,25 @@ def get_package_name(self, default=-1):
...
+        in the RPM's spec file.
+        :param default: If specified, this value will be returned if package name could not be found.
+        :return: The package name - Otherwise, the specified default.
+                If default is not passed in, an exception will be raised if the package name can't be found.
+        """
+        package_name = self.config.name  # This should be the package name for the RPM
+        if package_name:
+            return package_name
+
+        if default != -1:
```
sosiouxme, line 251: this looks weird / non-pythonic... default ought to be `None` and this
```suggestion
        if default is not None:
```
but i guess this is everywhere and we can fix that some other time when it annoys me :)

sosiouxme (approved):

doozerlib/rpmcfg.py
```diff
@@ -233,6 +234,25 @@ def get_package_name(self, default=-1):
...
+            return default
+
+        raise IOError(f'Missing package name in RPM doozer metadata')
```
sosiouxme, line 254: flake8 doesn't like this being an f-string.
Might be helpful to indicate which distgit has this problem though?

vfreex (approved):
Unit tests need to be fixed.

vfreex (changes requested):
Unit tests need to be fixed.

jupierce:
Fixed the test.

PR #426, opened 2021-06-04
Scan sources assembly

jupierce:
Makes get_latest_build and scan-sources assembly aware. As a side benefit, it also completely eliminates the need to clone upstream & downstream git repos. scan-sources can run in just a couple of minutes!

PR #427, opened 2021-06-09
Unify the way of extracting assembly names from release fields

vfreex:
Use `util.isolate_assembly_in_release` introduced by https://github.com/openshift/doozer/pull/426.


vfreex:
@jupierce I am going to merge this because this is very straightforward.

PR #428, opened 2021-06-11
ART-3034: Implement doozer images:rebase --force-yum-updates

vfreex:
`--force-yum-updates` will inject `yum update -y` in each stage. This ensures the component image will be able to override RPMs it is inheriting from its parent image using RPMs in the rebuild plashet.

If the option is not set, previously injected `yum update -y` commands
should be removed (in case we are injecting distgit-only repos).

Note if no repos are enabled for an image, we shouldn't inject `yum
update -y` otherwise the command will fail.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1697,7 +1697,7 @@ class ImageDistGitRepo(DistGitRepo):
             else:
                 self.logger.warning("Will not inject `USER 0` before `yum update -y` for the final build stage because `final_stage_user` is missing (or 0) in image meta."
                                     " If this build fails with `yum update -y` permission denied error, please set correct `final_stage_user` and rebase again.")
-            output.write(f"# {yum_update_line_flag}\n{yum_update_line}\n")
+            output.write(f"# {yum_update_line_flag}\n{yum_update_line}  # set final_stage_user in ART metadata if this fails\n")
             if build_stage == build_stage_num and final_stage_user:
                 output.write(f"# {yum_update_line_flag}\nUSER {final_stage_user}\n")
         output.seek(0)
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1361,11 +1361,8 @@ class ImageDistGitRepo(DistGitRepo):
                 if changed:
                     dfp.add_lines_at(entry, "RUN " + new_value, replace=True)
 
-    def update_distgit_dir(self, version, release, prev_release=None):
+    def update_distgit_dir(self, version, release, prev_release=None, force_yum_updates=False):
         ignore_missing_base = self.runtime.ignore_missing_base
-        # A collection of comment lines that will be included in the generated Dockerfile. They
-        # will be prefix by the OIT_COMMENT_PREFIX and followed by newlines in the Dockerfile.
-        oit_comments = []
 
         dg_path = self.dg_path
         with Dir(self.distgit_dir):
@@ -1646,10 +1643,10 @@ class ImageDistGitRepo(DistGitRepo):
                 'BUILD_RELEASE': release if release else '',
             }
 
+            df_fileobj = self._update_yum_update_commands(force_yum_updates, io.StringIO(df_content))
             with dg_path.joinpath('Dockerfile').open('w', encoding="utf-8") as df:
-                for comment in oit_comments:
-                    df.write("%s %s\n" % (OIT_COMMENT_PREFIX, comment))
-                df.write(df_content)
+                shutil.copyfileobj(df_fileobj, df)
+                df_fileobj.close()
 
             self._update_environment_variables(env_vars)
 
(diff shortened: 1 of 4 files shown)


doozerlib/distgit.py
```diff
@@ -1653,12 +1653,40 @@ def update_distgit_dir(self, version, release, prev_release=None):
...
+            # Remove "yum update -y" lines intead.
+            self.logger.warning("Will not inject \"yum updates -y\" for this image because no yum repos are enabled.")
+            force_yum_updates = False
+        if force_yum_updates:
+            self.logger.info("Injecting \"yum updates -y\" in each stage...")
+        df_path = self.dg_path.joinpath(dockerfile_path)
+        dfp = DockerfileParser(str(df_path))
+        df_lines = dfp.content.splitlines(False)
+
+        yum_update_line_flag = '__doozer=yum-update'
```
jupierce, line 1678: Clever!

jupierce (approved):

sosiouxme:
@jupierce @vfreex wait... why do we need this? blanket updates have been considered harmful for some time. you get into weird situations where something you've already updated conflicts with what you're trying to install.

also, i'm skeptical that they would even execute in all cases. not all images run as root user.

also, if we were going to yum update, we'd probably also want to yum clean all.

sosiouxme (changes requested):
@jupierce @vfreex wait... why do we need this? blanket updates have been considered harmful for some time. you get into weird situations where something you've already updated conflicts with what you're trying to install.

also, i'm skeptical that they would even execute in all cases. not all images run as root user.

also, if we were going to yum update, we'd probably also want to yum clean all.

vfreex:
@sosiouxme 

> why do we need this?

That's a good question. I would leave it to @jupierce :)

> i'm skeptical that they would even execute in all cases. not all images run as root user.
Right, in that case the build will fail. However it doesn't seem possible to change to ROOT user then change it back after the update.

> if we were going to yum update, we'd probably also want to yum clean all.
Good point.

jupierce:
@sosiouxme - the intent is to allow us to update RHEL dependencies in an assembly image. The flag would only be used in that context. Consider an assembly that is trying to update a RHEL RPM `foo` to version X in specific the machine-api-operator image v4.7.8. The basis image v4.7.8 has `foo` version Y. The "rebuild job" would do the following. 
1. Create an NVR list out of rpms installed in the basis image v4.7.8  (this contains Y).
2. Read assembly metadata to find dependencies to path in (this contains X).
3. Layer (2) over (1).
4. Create a plashet for `machine-api-operator` using (3). This plashet is only for a single image.
5. Run the build for the assembly with `--force-yum-updates` pointing `images:build` to ONLY this plashet.

The image build will run `yum update -y` but the only possible update will be X. The build output should be nearly identical to v4.7.8's machine-api-operator, but with X.

fwiw, I was also hoping to use the same mechanic to satisfy your "we need an assembly which is the same as v4.7.8 but with all RHEL dependencies updated". In this case, we would performing an an assembly specific rebuild using `--force-yum-updates` but _without_ first curating the image specific plashet.

jupierce:
@vfreex - +1 on `yum clean all` . For the USER issue, we need to add `USER 0` before the yum update. 
Since this is lossy, we could introduce metadata for images that are sensitive to it. For CI alignment PRs, where we have to change the user for a similar reason, we introduced https://github.com/openshift/ocp-build-data/blob/8e9703d0eeeaf5bd552c2780badff8d66fa27ba8/images/openshift-base-nodejs.yml#L12 . This is the only image that uses it (i.e. most of our images are running as UID 0 in CI). 

If you promote this field out of `ci_alignment` (https://github.com/openshift/ocp-build-data-validator/blob/4db9169a4f493f3cf803e5d1c78fd25102fb1a25/validator/schema/image_schema.py#L59) into the main metadata, we could solve this use case as well. Something like this in openshift-base-nodejs.yml

```yaml
...
from:  stream: nodejs
...
final_stage_user: 1001
```

The doozer CI alignment code and your new feature could then set `USER` in the *final stage* to this field (if not Missing) to the specified value. So your injection would look something like for the final stage:
```
FROM ....
USER 0
RUN yum update -y
RUN yum clean all
USER 1001
```

For non-final stages:
```
FROM ....
USER 0
RUN yum update -y
RUN yum clean all
```

If not running with `--force-yum-updates`, the injection for the final stage would just be:
```
FROM ....
USER 1001
```

vfreex:
@sosiouxme @jupierce 

Thanks for the review. I've updated my PR to inject `USER 0` and `yum clean all`.

We are injecting `yum update -y` next to the `FROM ` directive, before executing any directive from the source Dockerfile, in order to update packages inherited from the base image. But `images:streams gen-buildconfigs` is injecting to the end of Dockerfile. So the user field used in this PR is the user inherited from the parent image of the final stage, but the user in `images:streams gen-buildconfigs` is the user eventually embedded  in image manifest. So strictly speaking, they are different.

So, I would like to keep the top level `final_stage_user` separate from `content.source.ci_alignment.final_user`, and make `content.source.ci_alignment.final_user` default to `final_stage_user` in case we want to override. It doesn't seem correct to inject `yum update -y` to the end of each stage otherwise we may accidentally upgrade more than we needed. I could be wrong but let me know if it is fine.

Also I would prefer not to inject `USER 0` to the final stage if `final_stage_user` is missing. So if an image is building as non-root without `final_stage_user`, it will fail.

jupierce:
Opened validator PR to support this direction: https://github.com/openshift/ocp-build-data-validator/pull/34

doozerlib/distgit.py
```diff
@@ -1659,6 +1656,53 @@ def update_distgit_dir(self, version, release, prev_release=None):
...
+            # We need to change the current user to root.
+            # For non-final stages, it should be safe to inject "USER 0" before `yum update -y`.
+            # However for the final stage, injecting "USER 0" without changing the user back may cause unexpected behavior.
+            # Per https://github.com/openshift/doozer/pull/428#issuecomment-861795424, introduce a new metadata `final_stage_user` for images so we can switch the user back later.
+            if build_stage < build_stage_num or final_stage_user:
+                output.write(f"# {yum_update_line_flag}\nUSER 0\n")
+            else:
+                self.logger.warning("Will not inject `USER 0` before `yum update -y` for the final build stage because `final_stage_user` is missing (or 0) in image meta."
+                                    " If this build fails with `yum update -y` permission denied error, please set correct `final_stage_user` and rebase again.")
+            output.write(f"# {yum_update_line_flag}\n{yum_update_line}\n")
```
jupierce, line 1700:
```suggestion
            output.write(f"# {yum_update_line_flag}\n{yum_update_line}  # set final_stage_user in ART metadata if this fails\n")
```
Small hint for artists if they are only looking at brew logs?

vfreex: Yeah, a hint looks good to me.

vfreex:
Merging this based on the conversation with @jupierce

PR #431, opened 2021-06-11
Assembly lense

jupierce:
Implements the logic to load releases.yml and to have doozer interpret group and member metadata through assemblies defined in that file. 
Also eliminates some messy defaulting logic that used to be required before requiring that rpm metadata include package name in the its ocp-build-data file.

### DIFF
--- a/doozerlib/assembly.py
+++ b/doozerlib/assembly.py
@@ -1,0 +1,106 @@
+import copy
+
+from doozerlib.model import Model
+
+
+def merger(a, b):
+    """
+    Merges two, potentially deep, objects into a new one and returns the result.
+    Conceptually, 'a' is layered over 'b' and is dominant when
+    necessary. The output is 'c'.
+    1. if 'a' specifies a primitive value, regardless of depth, 'c' will contain that value.
+    2. if a key in 'a' specifies a list and 'b' has the same key/list, a's list will be appended to b's for c's list.
+       Duplicates entries will be removed and primitive (str, int, ..) lists will be returned in sorted order).
+    3. if a key ending with '!' in 'a' specifies a value, c's key-! will be to set that value exactly.
+    4. if a key ending with a '?' in 'a' specifies a value, c's key-? will be set to that value is 'c' does not contain the key.
+    """
+
+    if type(a) in [bool, int, float, str, bytes, type(None)]:
+        return a
+
+    c = copy.deepcopy(b)
+
+    if type(a) is list:
+        if type(c) is not list:
+            return a
+        for entry in a:
+            if entry not in c:  # do not include duplicates
+                c.append(entry)
+
+        if c and type(c[0]) in [str, int, float]:
+            return sorted(c)
+        return c
+
+    if type(a) is dict:
+        if type(c) is not dict:
+            return a
+        for k, v in a.items():
+            if k.endswith('!'):  # full dominant key
+                k = k[:-1]
+                c[k] = v
+            elif k.endswith('?'):  # default value key
+                k = k[:-1]
+                if k not in c:
+                    c[k] = v
+            else:
+                if k in c:
+                    c[k] = merger(a[k], c[k])
+                else:
+                    c[k] = a[k]
+        return c
+
+    raise IOError(f'Unexpected value type: {type(a)}: {a}')
+
+
+def group_for_assembly(releases_config: Model, assembly: str, group_config: Model):
+    """
+    Returns a group config based on the assembly information
+    and the input group config.
+    :param releases_config: A Model for releases.yaml.
+    :param assembly: The name of the assembly
+    :param group_config: The group config to merge into a new group config (original Model will not be altered)
+    """
+    if not assembly or not isinstance(releases_config, Model):
+        return group_config
+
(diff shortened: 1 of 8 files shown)


doozerlib/assembly.py
```diff
@@ -0,0 +1,138 @@
...
+                if k not in c:
+                    c[k] = v
+            else:
+                if k in c:
+                    c[k] = merger(a[k], c[k])
+                else:
+                    c[k] = a[k]
+        return c
+
+    raise IOError(f'Unexpected value type: {type(a)}: {a}')
```
vfreex, line 53: Seems a `TypeError` is more appropriate. 

```suggestion
    raise TypeError(f'Unexpected value type: {type(a)}: {a}')
```

jupierce: ack. Will be in next push.

doozerlib/assembly.py
```diff
@@ -0,0 +1,138 @@
...
+    None is returned.
+    """
+    if not assembly or not isinstance(releases_config, Model):
+        return None
+
+    target_assembly = releases_config.releases[assembly].assembly
+    if target_assembly.basis.brew_event:
+        return int(target_assembly.basis.brew_event)
+
+    return assembly_basis_event(releases_config, target_assembly.basis.assembly)
```
vfreex, line 124: If `target_assembly.basis.assembly` equals to `assembly` or is one of the assemblies this function has visited, this will lead to an infinity recursion.

Would like to prevent some human errors here by changing this function from recursion to iteration and raise an error if there's a cycle that would cause an infinity loop.

doozerlib/assembly.py
```diff
@@ -0,0 +1,138 @@
...
+    if target_assembly.basis.assembly:  # Does this assembly inherit from another?
+        # Recursive apply ancestor assemblies
+        meta_config = assembly_metadata_config(releases_config, target_assembly.basis.assembly, meta_type, distgit_key, meta_config)
+
+    config_dict = meta_config.primitive()
+
+    component_list = target_assembly.members[f'{meta_type}s']
+    for component_entry in component_list:
+        if component_entry.distgit_key == '*' or component_entry.distgit_key == distgit_key:
+            config_dict = merger(component_entry.metadata.primitive(), config_dict)
```
vfreex, line 105: The result of `merger` depends on the order of `component_entry` in the list. It seems to me we should merge component_entry for a specific distgit_key on the top of the wirdcard one in case of overriding.

jupierce: We might not ultimately need the mechanic, but I did intend for order to matter -- anticipating more exotic matching at some point. If it does prove to be overkill / confusing in the future, we can convert this to a dict and eliminate the expectation of ordering.

doozerlib/assembly.py
jupierce: Added a check.

doozerlib/metadata.py
```diff
@@ -381,11 +378,34 @@ def latest_build_list(pattern_suffix):
...
+                    return koji_api.getBuild(self.config['is'].nvr, strict=True)
+                else:
+                    raise IOError(f'"is" definition not yet understood: {self.config["is"]}')
+
             if not assembly:
                 # if assembly is '' (by parameter) or still None after runtime.assembly,
                 # we are returning true latest.
                 builds = latest_build_list('')
             else:
+                basis_event = assembly_basis_event(self.runtime.get_releases_config(), assembly=assembly)
```
vfreex, line 399:
```suggestion
                basis_event = self.runtime.assembly_basis_event if assembly == self.runtime.assembly else assembly_basis_event(self.runtime.get_releases_config(), assembly=assembly) 
```

jupierce: Is this for performance? Simpler testing? I doesn't seem like it would result in different values.

doozerlib/assembly.py
vfreex: ack

doozerlib/metadata.py
vfreex: This is unlikely to become a bottleneck, but I spent some time on this to figure out why reloading assembly_basis_event is needed (then I found it is just to support the `assembly` parameter rather than have special reasons). Changing this seems more readable to me, but it doesn't matter much.

doozerlib/assembly.py
```diff
@@ -0,0 +1,154 @@
...
+
+    if target_assembly.basis.assembly:  # Does this assembly inherit from another?
+        # Recursive apply ancestor assemblies
+        meta_config = assembly_metadata_config(releases_config, target_assembly.basis.assembly, meta_type, distgit_key, meta_config)
+
+    config_dict = meta_config.primitive()
+
+    component_list = target_assembly.members[f'{meta_type}s']
+    for component_entry in component_list:
+        if component_entry.distgit_key == '*' or component_entry.distgit_key == distgit_key:
```
vfreex, line 118: When I was testing assembly metadata pins the artifact NVR with `is:`, I didn't set `metadata`, then `metadata.primitive()` raised an error. Not sure if we need to allow None metadata here.

jupierce: Thanks! Fixed.

PR #432, opened 2021-06-15
Enable commit prefix in doozer metadata

jupierce:


vfreex:
/lgtm

Seems very straightforward to me.

The unit test failure is probably not related to this PR.

PR #439, opened 2021-06-18
ART-3050: full assembly awareness for plashet from-tags

vfreex:
When an artist is trying to refine an assembly with a basis event, existing assembly-specific builds for that assembly are not canonically part of the assembly until the artist pins then in the assembly metadata with "is". It follows that plashet should honor "is" when set for RPMs in the assembly metadata when constructing a repo.
    
Another change in behavior is that plashet should honor group "dependencies" defined in the assembly configuration. 431 takes care of determining what a group configuration should look like in light of a given assembly name on the command line. Unlike normal group.yml, assembly group definitions in releases.yml can add "dependencies" to the loaded models.

NOTE:
1. Plashet are able to create both rhel8 and rhel7 repos, but rpm nvrs defined in assembly config don't specify which repo they should go. I think we can assume assembly-specific nvrs either pinned by "is" or given by group dependencies are always for RHEL8. Let me know if this is wrong. Update: Now we have per rhel version dependencies in releases.yml.
2. It seems to me that we shouldn't allow "is" in a wildcard rpm member entry. Let me know if this is wrong. Update: This is now handled by `assembly_metadata_config()`
3. If the nvr pinned by "is" is embargoed but we are building an non-embargoed plashet, an error will be raised.
4. Remove `--brew-event` command line argument from `from-tags` subverb since we need to honor the event id defined in the basis.
5. This PR requires https://github.com/openshift/doozer/pull/431.
6. I used the following `releases.yml` for testing:

```yaml
releases:
  art1:
    assembly:
      basis: 
        brew_event: 39473862
      group:
        dependencies:
          rpms:
          - el8: cri-o-1.21.0-88.rhaos4.8.gitfd485de.el8
      members:
        rpms:
        - distgit_key: openshift-clients
          metadata:
            is:
              el8: openshift-clients-4.9.0-202106031118.p0.git.714295d.assembly.stream.el8
        - distgit_key: openshift-kuryr
          metadata:
            is:
              el8: openshift-kuryr-4.9.0-202105211327.p0.git.be00acd.el8
```

### DIFF
--- a/doozerlib/cli/plashet.py
+++ b/doozerlib/cli/plashet.py
@@ -17,6 +17,7 @@ import yaml
 from kobo.rpmlib import compare_nvr, parse_nvr
 from requests_kerberos import HTTPKerberosAuth
 
+from doozerlib.assembly import assembly_metadata_config
 from doozerlib.brew import get_build_objects
 from doozerlib.cli import cli
 from doozerlib.model import Missing
@@ -505,20 +506,11 @@ def from_tags(config: SimpleNamespace, brew_tag: Tuple[Tuple[str, str], ...], em
     runtime: Runtime = config.runtime
     runtime.initialize(mode="rpms", clone_source=False, clone_distgits=False, prevent_cloning=True)
     assembly = runtime.assembly
-    event = runtime.brew_event
     koji_proxy = runtime.build_retrying_koji_client()
     koji_proxy.gssapi_login()
+    event = runtime.brew_event
+    event_info = koji_proxy.getEvent(event)
     errata_proxy = xmlrpclib.ServerProxy(config.errata_xmlrpc_url)
-
-    if event:
-        if runtime.assembly_basis_event:
-            assert runtime.assembly_basis_event == event  # to be safe
-        event_info = koji_proxy.getEvent(event)
-    else:
-        # If none was specified, lock in an event so that there are no race conditions with
-        # packages changing while we run.
-        event_info = koji_proxy.getLastEvent()
-        event = event_info['id']
 
     # Gather up all nvrs tagged in the embargoed brew tags into a set.
     embargoed_tag_nvrs = set()
@@ -569,33 +561,24 @@ def from_tags(config: SimpleNamespace, brew_tag: Tuple[Tuple[str, str], ...], em
             el_version_match = re.search(r"rhel-(\d+)", tag)
             el_version = int(el_version_match[1]) if el_version_match else 0
             if el_version:  # Only honor pinned rpms if this tag is relevant to a RHEL version
-                # Honor pinned rpm nvrs specified in assembly.members.rpms
-                for component in runtime.releases_config.releases[runtime.assembly].assembly.members.rpms:
-                    nvr = component["is"][f"el{el_version}"]
+                # Honor pinned rpm nvrs pinned by "is"
+                for distgit_key, rpm_meta in runtime.rpm_map.items():
+                    meta_config = assembly_metadata_config(runtime.get_releases_config(), runtime.assembly, 'rpm', distgit_key, rpm_meta.config)
+                    nvr = meta_config["is"][f"el{el_version}"]
                     if not nvr:
                         continue
-                    if "*" in component.distgit_key:
-                        raise ValueError(f"You cannot pin an NVR in an assembly entry that has a wildcard in its distgit key ({component.distgit_key}).")
-                    rpm_meta = runtime.rpm_map.get(component.distgit_key)
-                    if not rpm_meta:
-                        raise ValueError(f"Distgit key {component.distgit_key} is unknown or excluded.")
                     nvre_obj = parse_nvr(str(nvr))
                     if nvre_obj["name"] != rpm_meta.rpm_name:
-                        raise ValueError(f"RPM {nvr} is pinned to assembly {runtime.assembly} for distgit key {component.distgit_key}, but its package name is not {rpm_meta.rpm_name}.")
-                    if isolate_assembly_in_release(nvre_obj["release"]) != runtime.assembly:
-                        raise ValueError(f"RPM {nvr} is pinned to assembly {runtime.assembly}, but that rpm is not a part of the assembly.")
-                    if nvre_obj["name"] in pinned_nvrs and pinned_nvrs[nvre_obj['name']] != nvr:
-                        raise ValueError(f"Cannot pin {nvr} because a different NVR {pinned_nvrs[nvre_obj['name']]} for the same component is already pinned.")
+                        raise ValueError(f"RPM {nvr} is pinned to assembly {runtime.assembly} for distgit key {distgit_key}, but its package name is not {rpm_meta.rpm_name}.")
                     pinned_nvrs[nvre_obj["name"]] = nvr
-                # assert pinned rpm nvrs specified in assembly.members.rpms exist in Brew then add them to plashet
                 if pinned_nvrs:
                     pinned_nvr_list = list(pinned_nvrs.values())
                     logger.info("Found %s NVRs pinned to the runtime assembly %s. Fetching build infos from Brew...", len(pinned_nvr_list), runtime.assembly)
                     pinned_builds = get_build_objects(pinned_nvr_list, koji_proxy)
                     missing_nvrs = [nvr for nvr, build in zip(pinned_nvr_list, pinned_builds) if not build]
                     if missing_nvrs:
-                        raise IOError(f"The following NVRs don't exist: {missing_nvrs}")
-                    # Pinned_builds should take precedence over very build in `latest_builds`
+                        raise IOError(f"The following NVRs pinned by 'is' don't exist: {missing_nvrs}")
+                    # Pinned_builds should take precedence over every build in `latest_builds`
                     for pinned_build in pinned_builds:
                         component = pinned_build["name"]
                         if component in component_builds and pinned_build["id"] != component_builds[component]["id"]:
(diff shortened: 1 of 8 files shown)


jupierce:
> Plashet are able to create both rhel8 and rhel7 repos, but rpm nvrs defined in assembly config don't specify which repo they should go. I think we can assume assembly-specific nvrs either pinned by "is" or given by group dependencies are always for RHEL8. Let me know if this is wrong.

Nice catch. I think this is a failing of the schema. I've changed the hackmd to use:
```yaml
dependencies:
  rpms:
  - el7: <nvr>
  - el8: <nr>
  - el9: <entries like this are not too far in the future>
```
This should allow you to find only NVRs relevant to the repo's RHEL version. I've also changed the schema for `is` when pinning member components:
```yaml
is:
  el7:  <nvr>
  el8:  <nvr>
```

`get_latest_build` in 431 has been updated to honor this. plashet could pass `el_target=#` in order for the right rpm build record to be returned.

For pinning image components, `is` remains a `nvr`:
```yaml
is:
  nvr: <nvr>
```

Let me know what you think.

> It seems to me that we shouldn't allow "is" in a wildcard rpm member entry. Let me know if this is wrong.

I agree, but I would consider it deep human error and not something I would worry too much about defending against.

> If the nvr pinned by "is" is embargoed but we are building an non-embargoed plashet, an error will be raised.

Perfect.

> Remove --brew-event command line argument from from-tags subverb since we need to honor the event id defined in the basis.

Great. Perfect place to rely on Runtime.

doozerlib/cli/plashet.py
```diff
@@ -501,12 +503,16 @@ def from_tags(config, brew_tag, embargoed_brew_tag, embargoed_nvr, signing_advis
...
     koji_proxy.gssapi_login()
     errata_proxy = xmlrpclib.ServerProxy(config.errata_xmlrpc_url)
 
     if event:
-        event = int(event)
+        if runtime.assembly_basis_event:
+            assert runtime.assembly_basis_event == event  # to be safe
         event_info = koji_proxy.getEvent(event)
     else:
         # If none was specified, lock in an event so that there are no race conditions with
```
jupierce, line 518: I wonder if we should simply move this logic into Runtime (i.e. if not part of the assembly and not specified on the command line, just getLastEvent). There no reason for doozer to run without a locked brew event.

doozerlib/cli/plashet.py
```diff
@@ -545,15 +550,85 @@ def from_tags(config, brew_tag, embargoed_brew_tag, embargoed_nvr, signing_advis
...
+                        continue
+                    if "*" in component.distgit_key:
+                        raise ValueError(f"You cannot pin an NVR in an assembly entry that has a wildcard in its distgit key ({component.distgit_key}).")
+                    rpm_meta = runtime.rpm_map.get(component.distgit_key)
+                    if not rpm_meta:
+                        raise ValueError(f"Distgit key {component.distgit_key} is unknown or excluded.")
+                    nvre_obj = parse_nvr(str(nvr))
+                    if nvre_obj["name"] != rpm_meta.rpm_name:
+                        raise ValueError(f"RPM {nvr} is pinned to assembly {runtime.assembly} for distgit key {component.distgit_key}, but its package name is not {rpm_meta.rpm_name}.")
+                    if isolate_assembly_in_release(nvre_obj["release"]) != runtime.assembly:
```
jupierce, line 585: I think we can take out this check. Consider that assemblies can inherit from one another.
If `art2` inherits from `art1`, and `art1` pins component X, then the isolated assembly name could be `art1` - which is perfectly valid. 
Also we could be pinning a particular component to an older, known-good, `stream` build to provide a custom hotfix.

doozerlib/cli/plashet.py
```diff
@@ -545,15 +550,85 @@ def from_tags(config, brew_tag, embargoed_brew_tag, embargoed_nvr, signing_advis
...
-
-        for build in latest_builds:
+            latest_builds = find_latest_builds(tagged_builds, assembly)
+            component_builds = {build["name"]: build for build in latest_builds}
+
+            el_version_match = re.search(r"rhel-(\d+)", tag)
+            el_version = int(el_version_match[1]) if el_version_match else 0
+            if el_version:  # Only honor pinned rpms if this tag is relevant to a RHEL version
+                # Honor pinned rpm nvrs specified in assembly.members.rpms
+                for component in runtime.releases_config.releases[runtime.assembly].assembly.members.rpms:
```
jupierce, line 573: This doesn't appear to take into account assemblies inheriting from other assemblies via `basis.assembly`. I'd suggest a new method in assembly.py like `assembly_member_config`.

doozerlib/cli/plashet.py
```diff
@@ -545,15 +550,85 @@ def from_tags(config, brew_tag, embargoed_brew_tag, embargoed_nvr, signing_advis
...
+                        raise ValueError(f"You cannot pin an NVR in an assembly entry that has a wildcard in its distgit key ({component.distgit_key}).")
+                    rpm_meta = runtime.rpm_map.get(component.distgit_key)
+                    if not rpm_meta:
+                        raise ValueError(f"Distgit key {component.distgit_key} is unknown or excluded.")
+                    nvre_obj = parse_nvr(str(nvr))
+                    if nvre_obj["name"] != rpm_meta.rpm_name:
+                        raise ValueError(f"RPM {nvr} is pinned to assembly {runtime.assembly} for distgit key {component.distgit_key}, but its package name is not {rpm_meta.rpm_name}.")
+                    if isolate_assembly_in_release(nvre_obj["release"]) != runtime.assembly:
+                        raise ValueError(f"RPM {nvr} is pinned to assembly {runtime.assembly}, but that rpm is not a part of the assembly.")
+                    if nvre_obj["name"] in pinned_nvrs and pinned_nvrs[nvre_obj['name']] != nvr:
```
jupierce, line 587: I also think this is acceptable in light of assembly inheritance. art1 may define an nvr and art2 can inherit from art1 and override `is` for a component specified by art1.

doozerlib/cli/plashet.py
```diff
@@ -545,15 +550,85 @@ def from_tags(config, brew_tag, embargoed_brew_tag, embargoed_nvr, signing_advis
...
+                        raise ValueError(f"Cannot pin {nvr} because a different NVR {pinned_nvrs[nvre_obj['name']]} for the same component is already pinned.")
+                    pinned_nvrs[nvre_obj["name"]] = nvr
+                # assert pinned rpm nvrs specified in assembly.members.rpms exist in Brew then add them to plashet
+                if pinned_nvrs:
+                    pinned_nvr_list = list(pinned_nvrs.values())
+                    logger.info("Found %s NVRs pinned to the runtime assembly %s. Fetching build infos from Brew...", len(pinned_nvr_list), runtime.assembly)
+                    pinned_builds = get_build_objects(pinned_nvr_list, koji_proxy)
+                    missing_nvrs = [nvr for nvr, build in zip(pinned_nvr_list, pinned_builds) if not build]
+                    if missing_nvrs:
+                        raise IOError(f"The following NVRs don't exist: {missing_nvrs}")
```
jupierce, line 597:
```suggestion
                        raise IOError(f"The following NVRs pinned by 'is' don't exist: {missing_nvrs}")
```

doozerlib/cli/plashet.py
```diff
@@ -545,15 +550,85 @@ def from_tags(config, brew_tag, embargoed_brew_tag, embargoed_nvr, signing_advis
...
+                    # Pinned_builds should take precedence over very build in `latest_builds`
```
jupierce, line 598:
```suggestion
                    # Pinned_builds should take precedence over every build in `latest_builds`
```

doozerlib/cli/plashet.py
```diff
@@ -545,15 +550,85 @@ def from_tags(config, brew_tag, embargoed_brew_tag, embargoed_nvr, signing_advis
...
+
+                # honor group dependencies
+                dep_nvrs = {parse_nvr(dep[f"el{el_version}"])["name"]: dep[f"el{el_version}"] for dep in runtime.group_config.dependencies.rpms if dep[f"el{el_version}"]}  # rpms for this rhel version listed in group dependencies; keys are rpm component names, values are nvrs
+                if dep_nvrs:
+                    dep_nvr_list = list(dep_nvrs.values())
+                    logger.info("Found %s NVRs defined in group dependencies. Fetching build infos from Brew...", len(dep_nvr_list))
+                    dep_builds = get_build_objects(dep_nvr_list, koji_proxy)
+                    missing_nvrs = [nvr for nvr, build in zip(dep_nvr_list, dep_builds) if not build]
+                    if missing_nvrs:
+                        raise IOError(f"The following NVRs don't exist: {missing_nvrs}")
```
jupierce, line 613:
```suggestion
                        raise IOError(f"The following group dependency NVRs don't exist: {missing_nvrs}")
```

jupierce (changes requested):

doozerlib/cli/plashet.py
jupierce: Actually, I can't think of any reason not to put `is` in metadata. https://github.com/openshift/ocp-build-data-validator/pull/35/commits/2c5af0845b8c21245adf86b4eb85fb8560456700

This should let you call:
```
meta_config = assembly_metadata_config(runtime.get_releases_config(), runtime.assembly, distgit_key, Model())
if meta_config['is']: ....
```

This will pick up any inherited value.

vfreex: Yeah, that sounds easier and we can reuse the logic in different places.

doozerlib/cli/plashet.py
vfreex: I am going to move it to Runtime.

doozerlib/cli/plashet.py
vfreex: By moving "is" into metadata, I think rpm meta will take care of the assembly inheritance so the rpm name will never repeat. Thoughts?

doozerlib/cli/plashet.py
vfreex: Makes sense. I will remove the check.

vfreex:
@jupierce I've addressed all issues you mentioned. Please have a look again.

doozerlib/cli/plashet.py
jupierce: Agreed.

jupierce (approved):

vfreex:
@jupierce I pushed another commit to extract some logic in `from-tags` subverb into a separate module so it can be reused in `for-assembly` verb. Could you have a look again?

vfreex:
Merging so that future development can base on this.

PR #442, opened 2021-06-21
[ART-3072] Fix version extraction

joepvd:
With a new distgit repository, this would explode. Allow for a not
existing `version` label.


vfreex (commented):
This PR itself LGTM, but @joepvd did you check if the caller of this function handles `None` version correctly?

thiagoalessio (approved):
/lgtm

PR #446, opened 2021-06-24
Correct click setup

joepvd:
Since click 8.0, it is a fatal error to have multiple=True and a Default
that is not a list. This allows for a new click.

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -1648,7 +1648,7 @@ def config_mode(runtime, mode, push, message):
 
 
 @cli.command("config:print", short_help="View config for given images / rpms")
-@click.option("-n", "--name-only", default=False, is_flag=True, multiple=True,
+@click.option("-n", "--name-only", default=False, is_flag=True,
               help="Just print name of matched configs. Overrides --key")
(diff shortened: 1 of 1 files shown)


thiagoalessio (approved):
/lgtm

doozerlib/cli/__main__.py
```diff
@@ -1648,7 +1648,7 @@ def config_mode(runtime, mode, push, message):
 
 
 @cli.command("config:print", short_help="View config for given images / rpms")
-@click.option("-n", "--name-only", default=False, is_flag=True, multiple=True,
+@click.option("-n", "--name-only", default=False, is_flag=True,
```
thiagoalessio, line 1651: alternatively, if we wanted to keep `multiple=True`, we can just change the default value:
```suggestion
@click.option("-n", "--name-only", default=[], is_flag=True, multiple=True
```

thiagoalessio: ha! Sid was faster: https://github.com/openshift/doozer/pull/443/files

joepvd:
Yep, missed Sid's PR. Happy to go with that one :)

https://github.com/openshift/doozer/pull/443

PR #450, opened 2021-06-25
[ART-3088] Remove obsolete operator-metadata commands

thiagoalessio:
No longer needed once 4.5 goes away


joepvd:
🎉 

/cc @vikaslaad

vikaslaad:
We should still keep it for some time even after 4.5 is EOL, lets not merge this until I confirm from BU.

vfreex:
I would prefer not to merge this until we get clear notice that 4.5 is EOL.
/hold

sosiouxme:
@vikaslaad is there any chance we would actually update and ship an OLM operator for 4.5 now? the payload, maybe, but operators? that would basically require shipping an actual image...

PR #452, opened 2021-06-28
Allow --repo to take precedence over --repo-type defaults

jupierce:
In order to image rebuilds for assemblies to function as designed, we need to be able to point brew to one more more explicit repos instead of depending on the defaults implied by `--repo-type`. Previously, `--repo` would be one element added to the defaults, but it will now be treated as the complete list of repos to send to brew.

Related to: https://issues.redhat.com/browse/ART-3091


vfreex (approved):
/lgtm

PR #453, opened 2021-06-28
Add additional event-safe methods for koji

jupierce:
Now that doozer's runtime is always event locked, we need to expand the list of "event-safe" koji methods that can be called at runtime. This simply means the methods cannot be and do not need to be event constrained.


jupierce:
fyi @vfreex

PR #454, opened 2021-06-29
Fix golang image build

vfreex:
Use tag based `get_latest_build` because golang builder images don't
follow the version numbering scheme like normal OCP images.


joepvd (approved):
/lgtm

PR #457, opened 2021-06-30
[ART-3092] Allow gen-payload to honor metadata.is

jupierce:
metadata.get_latest_build honored `is` already, so leveraging it for now


vfreex:
/lgtm
Needs to fix the unit tests.

jupierce:
`meta.get_latest_build` already has coverage, so I think we can drop this for now.

PR #458, opened 2021-06-30
[ART-2687] Expose brew event and brew tag to modification scripts

joepvd:


doozerlib/distgit.py
```diff
@@ -2245,7 +2245,11 @@ def _run_modifications(self):
                     "component_name": self.metadata.distgit_key,
                     "kind": "Dockerfile",
                     "content": new_dockerfile_data,
-                    "set_env": {"PATH": path},
+                    "set_env": {
+                        "PATH": path,
+                        "BREW_EVENT": f'{self.runtime.brew_event}',
+                        "BREW_TAG": f'{self.runtime.branch}-candidate'
```
vfreex, line 2251: `BREW_TAG` should be `self.metadata.default_brew_tag()` because in case of non-stream assembly build, the actual brew tag will be `$branch-hotfix`.

vfreex (changes requested):
Only 1 issue, otherwise LGTM.

doozerlib/distgit.py
joepvd: Thanks!

jupierce (approved):

PR #462, opened 2021-07-06
Set default assembly name to "stream"

joepvd:
I'd expect this command to produce a working and sensible link:

```
⫸  doozer --group openshift-4.7 config:read-group --yaml repos.rhel-8-server-ose-rpms-embargoed.conf.baseurl.x86_64 2>/dev/null
http://download.lab.bos.redhat.com/rcm-guest/puddles/RHAOS/plashets/4.7-el8/test/building-embargoed/x86_64/os
```

With `--assembly stream` to the global options this becomes clickable.
This sets the default assembly name to the one that is used in the
common case.

### DIFF
--- a/doozerlib/cli/__init__.py
+++ b/doozerlib/cli/__init__.py
@@ -50,7 +50,7 @@ def print_version(ctx, param, value):
               help="Username for rhpkg. Env var: DOOZER_USER")
 @click.option("-g", "--group", default=None, metavar='NAME[@commitish]',
               help="The group of images on which to operate. Env var: DOOZER_GROUP")
-@click.option("--assembly", metavar="ASSEMBLY_NAME", default='test',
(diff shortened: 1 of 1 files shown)


vfreex:
No, defaulting to `test` is by design, see https://issues.redhat.com/browse/ART-2940.

vfreex:
In Justin's design, only automated incremental builds belong to the `stream` assembly. If you invoke Doozer from the command line it will default to `test` assembly unless you specify the `--assembly` option.

The assembly name is part of the repo url. Builds for different assemblies will have different paths. So explicitly passing `--assembly=<name>` is reasonable.

jupierce:
If defaulting to `test` is counterintuitive, we could also just make the `--assembly` command line argument required.

joepvd:
Thanks for providing context! It makes sense to me now :)

PR #463, opened 2021-07-07
olm-bundles: make --force work, also not require user config

sosiouxme:
`--force` on this now will actually rebuild.
spaced out the `rhpkg{}clone` pattern which comes out `rhpkgclone` in my normal usage.


vfreex (approved):
/lgtm

PR #467, opened 2021-07-13
Scan sources fix assembly

jupierce:
Together, these changes should stop unnecessary rebuilds being suggested by scan-sources when a different assembly builds and RPM or image. 
- The metadata digest check is now performed relative to the image's actual build commit - not HEAD.
- The RPM check will ignore RPMs which are part of the group and are the latest NVR associated with the current assembly.


vfreex (approved):
/lgtm

PR #468, opened 2021-07-14
Fix parent stream resolution with a basis event

vfreex:
The tag name doesn't contain the component name.


jupierce (approved):
Nice catch!

PR #469, opened 2021-07-14
Always build against -candidate target then tag into -hotfix if needed

vfreex:
When building for non-stream assembly, tag the build into `-hotfix`
after build completes.

Also writes build NVRs to record.log so that our Jenkins job like
`rebuild` won't have to query Brew for NVRs by task IDs.


jupierce (approved):
Great! Thanks!

PR #473, opened 2021-07-15
Fix images:build fails to get the build NVR by task ID

vfreex:
This issue is introduced by https://github.com/openshift/doozer/pull/469.

Unlike rpm build, koji_api.listBuilds(taskID=...) doesn't support image build. I believe it is a koji bug.
For now, let's use a different approach.


thiagoalessio (approved):
/lgtm

PR #475, opened 2021-07-15
force-yum-updates updates final stage only

jupierce:
The original intent here was to allow an artist to pin an assembly group
or assembly image dependency in releases.yml and have builder stages
in the Dockerfile install those updated packages.

In practice, builder images are diverse and https://coreos.slack.com/archives/GDBRP5YJH/p1626358901421900?thread_ts=1626187378.370800&cid=GDBRP5YJH
shows at least one instance where the builder image could not be updated
with "yum -y update" even with access to all of rhel8. The reason being
that a -devel package was installed in the builder and that package was not
available in RHEL.

So the plan is, for now, to eliminate the ability to update a builder
image using assembly dependency pinning. We could accomplish this in the future
by only enabling the rebuild override repo for builder stages (i.e.
exposing ONLY the pinned dependencies indicated by the artist (instead
of all of RHEL via the image's group.yml repos). Since this could be tricky to accomplish elegantly (e.g
it would seem to require disabling non-override repos in build stages
prior to running yum -y update), we will instead move ahead without
this feature.

If an artist truly needs to override a specific package in a builder
image for an assembly, they can rebuild the builder images with the
desired package and then pin that builder image as an override in the
assembly definition.


doozerlib/distgit.py
```diff
@@ -1762,18 +1762,25 @@ def _update_yum_update_commands(self, force_yum_updates: bool, df_fileobj: io.Te
...
-            # If the curent user inherited from the base image for this stage is not root, `yum update -y` will fail.
-            # We need to change the current user to root.
-            # For non-final stages, it should be safe to inject "USER 0" before `yum update -y`.
-            # However for the final stage, injecting "USER 0" without changing the user back may cause unexpected behavior.
-            # Per https://github.com/openshift/doozer/pull/428#issuecomment-861795424, introduce a new metadata `final_stage_user` for images so we can switch the user back later.
-            if build_stage < build_stage_num or final_stage_user:
+
+            if build_stage != build_stage_num:
+                # If this is not the final stage, ignore this FROM
+                continue
```
sosiouxme, line 1768: [DockerfileParser.add_lines](https://github.com/containerbuildsystem/dockerfile-parse/blob/9f22a64d8cd037cb191aca69dc36bba23a831aab/dockerfile_parse/parser.py#L735) ought to simplify this. Also can't we remove the bit above checking for rhel7?

vfreex: We still need `yum install -y yum-plugin-priorities` for RHEL7, but the `if` check doesn't seem to be needed anymore.

vfreex:
Tested this PR with https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/hack/job/yuxzhu/job/aos-cd-jobs/job/build%252Frebuild/54/

vfreex:
@jupierce I think we can remove  the `if cat /etc/redhat-release | grep 'release ...'` check from injected commands to make the error more explicit. Otherwise LGTM.

vfreex:
@jupierce I commit a change to this PR to remove the RHEL version check from injected `yum update` commands because it is not needed if we only update the final stage. Let me know if you have concerns or have a better solution.

vfreex (approved):
/lgtm

PR #477, opened 2021-07-17
[ART-3145] Allow git "branch" in image metadata to be hash

jupierce:
During custom assembly rebuilds, it is important to be able to specify specific git commit hashes instead of just branch names. This allows doozer to work with commit hashes instead of just branch names.


vfreex (approved):
/lgtm

PR #481, opened 2021-07-20
release:gen-payload: Don't tag builds for standard releases into -hot…

vfreex:
…fix tags

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -219,7 +219,7 @@ class PayloadGenerator:
         brew_latest_builds = []
         for image_meta in payload_images:
             latest_build: ImageMetadata = image_meta.get_latest_build()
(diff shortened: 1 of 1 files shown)


vfreex:
@jupierce I am not sure this behavior is by design. Let me know if this is wrong.

jupierce:
@vfreex - If we continue to think about using hotfix for now simply as a means to prevent garbage collection, I don't think tagging stream builds is invalid. Hopefully, it is redundant since the stream build should release through the errata - but what if it doesn't?
Consider the scenario you brought up in another pull - where we define 4.9.22 and _it does not ship_. Then we define 4.9.23 to inherits from 4.9.22.
Without `-hotfix` tags, it is possible that images from 4.9.22 could be gc'd. Obviously, it is unlikely, but I would feel better for now if everything was safe from GC.

vfreex:
@jupierce Ack. Thanks.

PR #483, opened 2021-07-26
distgit: bundles can specify image-refs anywhere

sosiouxme:
having image-references mixed in with manifests causes problems in other environments. [ref](https://coreos.slack.com/archives/CB95J6R4N/p1627331310041400?thread_ts=1627327066.027000&cid=CB95J6R4N)

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -1794,11 +1794,15 @@ class ImageDistGitRepo(DistGitRepo):
 
     def _get_csv_file_and_refs(self, csv_config):
         gvars = self.runtime.group_config.vars
-        subdir = csv_config.get('bundle-dir', f'{gvars["MAJOR"]}.{gvars["MINOR"]}')
-        manifests = os.path.join(self.distgit_dir, csv_config['manifests-dir'], subdir)
+        bundle_dir = csv_config.get('bundle-dir', f'{gvars["MAJOR"]}.{gvars["MINOR"]}')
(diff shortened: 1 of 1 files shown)


doozerlib/distgit.py
```diff
@@ -1797,9 +1797,14 @@ def _get_csv_file_and_refs(self, csv_config):
         subdir = csv_config.get('bundle-dir', f'{gvars["MAJOR"]}.{gvars["MINOR"]}')
         manifests = os.path.join(self.distgit_dir, csv_config['manifests-dir'], subdir)
 
-        refs = os.path.join(manifests, 'image-references')
-        if not os.path.isfile(refs):
-            raise DoozerFatalError('{}: file does not exist: {}'.format(self.metadata.distgit_key, refs))
+        refs = None
+        ref_candidates = [os.path.join(path, 'image-references') for path in [csv_config['manifests-dir'], subdir, manifests]]
```
thegreyd, line 1801: these would need to be updated to absolute paths. I cleaned it up a bit at https://github.com/thegreyd/doozer/commit/60ffb2d9cbc284d3e88d1d0415b2ddbd8847c46b

thegreyd: Also are we changing the behavior for finding csvs as well?

sosiouxme: nope. just the single file.

sosiouxme: cp'ed that commit

vfreex (approved):
/lgtm

PR #495, opened 2021-08-09
Fix a bug where upstream git commit was incorrectly pinned for stream…

vfreex:
… assembly

Commit 0917797db7fce1b6e4ea55aefdce0e55b819cc46 introduced a new
behavior to prevent from rebuilding images/rpms off their latest commits
for non-stream assemblies.

However, assembly related fields in `class Runtime` are initialized too late,
which results in stream assembly builds are also pinned to their latest
upstream commits.

This PR moves up the code for initializing assembly related fields and
explicitly checks `assembly_basis_event` to decide whether pinning a commit is needed or not.


joepvd (approved):
🚀

PR #498, opened 2021-08-13
[ART-3069] olm-bundle assembly awareness

vfreex:
1. Currently `olm-bundle:rebase` and `olm-bundle:rebase-and-build` verbs
   require explicitly passing operator NVRs via command line arguments.
This PR keeps this behavior but also allows to run those verbs without
operator NVRs. In this case, operator NVRs selected by the runtime
assembly will be used for bundle builds.

2. Improve debug and troubleshoot capability by adding more logs and writing build related info to `record.log`.

3. Patch `images:build` to write `has_olm_bundle` field to `record.log`. This makes it easier to run `olm-bundle` verbs from Jenkins jobs by filtering image builds with OLM bundles.

4. Honor `--upcycle` and `--dry-run` options for better testing.

### DIFF
--- a/doozerlib/brew.py
+++ b/doozerlib/brew.py
@@ -441,6 +441,7 @@ class KojiWrapper(koji.ClientSession):
         'listArchives',
         'listRPMs',
         'getPackage',
+        'getPackageID',
         'listTags',
         'gssapi_login',
         'sslLogin',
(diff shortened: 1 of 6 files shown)


jupierce (approved):
Looks like a solid plan to me. Thanks @vfreex!

PR #496, opened 2021-08-11
[ART-3249] Bundles: Ensure search-and-replace is effective

joepvd:

PR #499, opened 2021-08-17
[ART-3249] Log ineffective search-and-replace

joepvd:
Follow-up on https://github.com/openshift/doozer/pull/496. Let's log
and work with upstream before erroring out.


thiagoalessio (approved):
/lgtm

PR #515, opened 2021-09-15
fix traceback when a wip/disabled operator is given

thegreyd:


sosiouxme:
i'm not sure it's wise to enable bundle commands for disabled/wip. seems like that would have us attempting builds for stuff we know will fail (or hasn't been built at all yet).

the reason we disable operators at branch cut is because they don't rebase successfully, and that blocks any other builds. once the operator rebase is working, i think it's ok to re-enable it, even if the bundle build ultimately fails. bundle builds don't block anything (except releases).

sosiouxme:
on the other hand, looks like `olm_bundle` now is being invoked with specific operators that got built. maybe it would be good in that case for the job to supply the `--load-wip` and `--load-disabled` flags to doozer. i just don't want doozer to default to that if we tell it to build everything.

thegreyd:
I've removed the changes, and this PR now just fixes the traceback issue. Merging

PR #518, opened 2021-09-16
release_gen_payload: fix default_imagestream_base_name usage

sosiouxme:
this method changed signature. scan_sources was using the previous
signature and began silently failing, causing it not to detect that
rhcos has an update.

the usage of rhcos.latest_rhcos_build_id also changed out from under
scan_sources - fixed.

### DIFF
--- a/doozerlib/cli/scan_sources.py
+++ b/doozerlib/cli/scan_sources.py
@@ -265,24 +265,19 @@ def _detect_rhcos_status(runtime, kubeconfig) -> list:
     for arch in runtime.arches:
         for private in (False, True):
             name = f"{version}-{arch}{'-priv' if private else ''}"
-            try:
-                tagged_mosc_id = _tagged_mosc_id(kubeconfig, version, arch, private)
-                latest_rhcos_id = _latest_rhcos_build_id(runtime, version, arch, private)
-                status = dict(name=name)
-                if not latest_rhcos_id:
-                    status['changed'] = False
-                    status['reason'] = "could not find an RHCOS build to sync"
-                elif tagged_mosc_id == latest_rhcos_id:
-                    status['changed'] = False
-                    status['reason'] = f"latest RHCOS build is still {latest_rhcos_id} -- no change from istag"
-                else:
-                    status['changed'] = True
-                    status['reason'] = f"latest RHCOS build is {latest_rhcos_id} which differs from istag {tagged_mosc_id}"
-                statuses.append(status)
-            except Exception as ex:
-                # don't let flakiness in rhcos lookups prevent us from scanning regular builds;
-                # if anything else changed it will sync anyway.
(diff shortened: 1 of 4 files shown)


sosiouxme:
/hold
before I merge this, we should probably understand why this is the output now:
```
rhcos:
-   changed: false
    name: 4.8-x86_64
    reason: latest RHCOS build is still 48.84.202109161436-0 -- no change from istag
-   changed: true
    name: 4.8-x86_64-priv
    reason: latest RHCOS build is 48.84.202109161436-0 which differs from istag 48.84.202107301701-0
-   changed: false
    name: 4.8-ppc64le
    reason: latest RHCOS build is still 48.84.202109161620-0 -- no change from istag
-   changed: true
    name: 4.8-ppc64le-priv
    reason: latest RHCOS build is 48.84.202109161620-0 which differs from istag 48.84.202107302004-0
-   changed: false
    name: 4.8-s390x
    reason: latest RHCOS build is still 48.84.202109161304-0 -- no change from istag
-   changed: true
    name: 4.8-s390x-priv
    reason: latest RHCOS build is 48.84.202109161304-0 which differs from istag 48.84.202107301721-0
```

is build-sync not updating the `-priv` imagestreams? i believe this would flag RHCOS as perpetually requiring an update.

sosiouxme:
/hold cancel
build-sync indeed was not syncing `-priv` streams. that got fixed, we're ready for this. (would be nice to have reviews)

doozerlib/cli/scan_sources.py
```diff
@@ -302,10 +302,10 @@ def _tagged_mosc_id(kubeconfig, version, arch, private) -> str:
     return stdout if stdout else None
 
 
-def _latest_rhcos_build_id(version, arch, private) -> str:
+def _latest_rhcos_build_id(runtime, version, arch, private) -> str:
     """wrapper to return None if anything goes wrong, which will be taken as no change"""
     try:
-        return rhcos.latest_rhcos_build_id(version, arch, private)
+        return rhcos.RHCOSBuildFinder(runtime, version, arch, private).latest_rhcos_build_id()
     except Exception:
```
jupierce, line 309: Seems like we could narrow this catch. It hid one bug and could hide another.

jupierce (approved):
One thought, but otherwise lgtm.

doozerlib/cli/scan_sources.py
sosiouxme: tough to narrow, there's a lot that can go wrong during the attempt. i guess we could start with obvious and add more as we encounter them.

i also would like to get around to unit testing this specifically...

sosiouxme:
i pushed the original commit directly to master, but the followup commits improve the error handling and unit testing so we're less likely to see this again.

PR #520, opened 2021-09-27
[ART-3279] Mirror arm64 builder and base images

Ximinhan:
sync arm64 builder and base images for to CI registry

### DIFF
--- a/doozerlib/cli/images_streams.py
+++ b/doozerlib/cli/images_streams.py
@@ -66,6 +66,7 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
     for upstream_entry_name, config in upstreaming_entries.items():
         if config.mirror is True or user_specified:
             upstream_dest = config.upstream_image
+            mirror_arm = False
             if upstream_dest is Missing:
                 raise IOError(f'Unable to mirror {upstream_entry_name} since upstream_image is not defined')
 
@@ -74,6 +75,8 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
             # for transforming upstream_image_base to upstream_image.
             if config.upstream_image_base is not Missing:
                 upstream_dest = config.upstream_image_base
+                if "aarch64" in runtime.arches:
+                    mirror_arm = True
 
             brew_image = config.image
             brew_pullspec = runtime.resolve_brew_image_url(brew_image)
@@ -81,9 +84,16 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
             if only_if_missing:
                 check_cmd = f'oc image info {upstream_dest}'
                 rc, check_out, check_err = exectools.cmd_gather(check_cmd)
-                if rc == 0:
-                    print(f'Image {upstream_dest} seems to exist already; skipping because of --only-if-missing')
-                    continue
+                if mirror_arm:
+                    check_cmd_arm = f'oc image info {upstream_dest}-arm64'
+                    rc_arm, check_out, check_err = exectools.cmd_gather(check_cmd_arm)
+                    if rc == 0 and rc_arm == 0:
+                        print(f'Image {upstream_dest} and {upstream_dest}-arm64 seems to exist already; skipping because of --only-if-missing')
+                        continue
+                else:
+                    if rc == 0:
+                        print(f'Image {upstream_dest} seems to exist already; skipping because of --only-if-missing')
+                        continue
 
             if live_test_mode:
                 upstream_dest += '.test'
@@ -97,7 +107,8 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
             else:
                 exectools.cmd_assert(cmd, retries=3, realtime=True)
 
-            if "aarch64" in runtime.arches:
+            # mirror arm64 builder and base images for CI
+            if mirror_arm:
                 arm_cmd = f'oc image mirror --filter-by-os linux/arm64 {brew_pullspec} {upstream_dest}-arm64'
                 if runtime.registry_config_dir is not None:
                     arm_cmd += f" --registry-config={get_docker_config_json(runtime.registry_config_dir)}"
--- a/doozerlib/cli/images_streams.py
+++ b/doozerlib/cli/images_streams.py
@@ -96,6 +96,13 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
                 print(f'For {upstream_entry_name}, would have run: {cmd}')
             else:
                 exectools.cmd_assert(cmd, retries=3, realtime=True)
+            
+            if runtime.group_config.get('arches', ['aarch64']):
+                arm_cmd = f'oc image mirror --filter-by-os linux/arm64 {brew_pullspec} {upstream_dest}-arm64'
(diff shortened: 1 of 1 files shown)


doozerlib/cli/images_streams.py
```diff
@@ -97,6 +97,13 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
             else:
                 exectools.cmd_assert(cmd, retries=3, realtime=True)
 
+            if "aarch64" in runtime.arches:
+                arm_cmd = f'oc image mirror --filter-by-os linux/arm64 {brew_pullspec} {upstream_dest}-arm64'
```
LorbusChris, line 101: These images haven't been transformed to include the CI repos, which will have to be added as another layer in another build step on the arm build farm.
I therefore wonder whether it makes to suffix them with ´-art-arm64´ here, and use the ´-arm64´ suffix for the transformed (i.e. CI repo layer added) images?

Ximinhan: oh, we only need apply this on those items have transform value ? If true I can add a  judgment condition

LorbusChris: What images would be mirrored with this step? I think we only need the base (could probably use that without the transform), and the builder, which will require the transform layer

Ximinhan: according to the [request](https://issues.redhat.com/browse/ART-3279) and[ luke's comment ](https://issues.redhat.com/browse/ART-3279?focusedCommentId=19074383&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-19074383) I think I will just mirror builder and base with `-arm64` suffix.

Ximinhan: according to the request and[ luke's comment ](https://issues.redhat.com/browse/ART-3279?focusedCommentId=19074383&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-19074383) I think I will just mirror builder and base with `-arm64` suffix.

LorbusChris: @sosiouxme since we'll still need to apply the transforms to the images and commit them back to the registry, wdyt about using the `-art-arm64` suffix here to distinguish them from the `-arm64` images which would include the transform layer? Could you also point me towards the definition of that transform so I can add a build job to apply it on the Prow side?

sosiouxme: if you look at e.g. https://github.com/openshift/ocp-build-data/blob/f883a68216730db80bb49b5fd69f899861087f50/streams.yml#L30-L38 we have something of a convention around this already where we push to a name with `.art` on the end, and then there's a build to transform it into what actually gets used (which is what we put in our alignment PRs so CI will point at them).

so I think we can continue that here by simply adding the suffix. so e.g. 
`upstream_image_base: registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-{MAJOR}.{MINOR}.art`
- means we will push the ARM version to `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9.art-arm64`
- which will then require a transform into `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9-arm64`
- according to `upstream_image: registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-{MAJOR}.{MINOR}`

x86 CI is configured to use `registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.9` and so arm64 CI should be configured to use that with the `-arm64` suffix (although maybe with the ARM64 cluster registry, will leave it to you how that last mile works).

sound good?

as far as the transforms... the transform in this example is
`transform: rhel-8/golang` which can be found in the [ci_transforms/rhel-8/golang](https://github.com/openshift/ocp-build-data/blob/openshift-4.9/ci_transforms/rhel-8/golang/) dir - it's just an openshift build with the base as the FROM, using that Dockerfile. I think that should translate directly into an ARM64 cluster build? I'm noticing though that the [base-repos transform](https://github.com/openshift/ocp-build-data/blob/openshift-4.9/ci_transforms/rhel-8/base-repos/Dockerfile) doesn't seem to be actually used any more, instead Test Platform is configured to add the repos some other way... maybe @jupierce could offer guidance?

doozerlib/cli/images_streams.py
```diff
@@ -97,6 +107,14 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
             else:
                 exectools.cmd_assert(cmd, retries=3, realtime=True)
 
+            # mirror arm64 builder and base images for CI
+            if mirror_arm:
+                arm_cmd = f'oc image mirror --filter-by-os linux/arm64 {brew_pullspec} {upstream_dest}-arm64'
```
sosiouxme, line 112: I'm was thinking this will fail for images that have no ARM arch (rhel-7 golang). However as it turns out, `oc image mirror` seems to just filter out missing arches (as long as the manifest is there) regardless of specifying `--skip-missing`. So this is fine. A comment that mentions this would be a good idea.

sosiouxme (changes requested):
This looks great, I just have a couple tweaks.

doozerlib/cli/images_streams.py
Ximinhan: I didn't see skip-missing will check specific arch, if there is no such arch it will skip.
https://github.com/openshift/oc/blob/master/pkg/cli/image/mirror/mirror.go#L500
https://github.com/openshift/oc/blob/master/pkg/cli/image/mirror/mirror.go#L538

doozerlib/cli/images_streams.py
LorbusChris: thanks for the details.
cc'ing @alvaroaleman @stevekuznetsov for their Prow expertise. Maybe you could sanity check this approach and point out any potential issues that you see with it, i.e. convert the transforms into buildconfigs running on a Prow cluster (specifically, arm01). Are we loosing any build or image-ref replacement automation with this?

doozerlib/cli/images_streams.py
```diff
@@ -109,10 +109,13 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
 
             # mirror arm64 builder and base images for CI
             if mirror_arm:
+                # oc image mirror will filter out missing arches (as long as the manifest is there) regardless of specifying --skip-missing
                 arm_cmd = f'oc image mirror --filter-by-os linux/arm64 {brew_pullspec} {upstream_dest}-arm64'
```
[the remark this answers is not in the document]
Ximinhan: I think the upstream_dest already include the art suffix if need, like this https://github.com/openshift/ocp-build-data/blob/openshift-4.9/streams.yml#L63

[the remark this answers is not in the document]
LorbusChris: Ahh yeah it looks like like `{upsteam_dest}` resolves to `config.upstream_image_base` here if that exists, so this should indeed be good, thanks!

doozerlib/cli/images_streams.py
sosiouxme: right, just thought that would make a helpful comment in the code :)

sosiouxme (approved):
thanks!

PR #521, opened 2021-09-27
ART-3107 Change operator channel names to stable (4.10)

thegreyd:
https://issues.redhat.com/browse/ART-3107

Test run:
./doozer --assembly=test -g openshift-4.10 olm-bundle:rebase-and-build --force sriov-network-operator-container-v4.10.0-202109150958.p0.git.5e1261d.assembly.stream

Changes visible at
- https://pkgs.devel.redhat.com/cgit/containers/sriov-network-operator-bundle/tree/metadata/annotations.yaml?h=rhaos-4.10-rhel-8
- https://pkgs.devel.redhat.com/cgit/containers/sriov-network-operator-bundle/tree/Dockerfile?h=rhaos-4.10-rhel-8

PR #523, opened 2021-10-07
ART-3107 Differentiate between default and extra channel stable mode

thegreyd:
https://issues.redhat.com/browse/ART-3107
Follow up from https://github.com/openshift/doozer/pull/521

We need 3 modes for group.yml key `operator_channel_stable` as per OCP version:
- for ocp versions we want no change in, the key will simply be absent
- for ocp versions we want to have "stable" as only a regular publish channel - "extra"
- for ocp versions we want to have "stable" as a default publish channel - "default"


sosiouxme:
:shipit:

PR #524, opened 2021-10-12
[ART-3109] Use container.yaml to apply floating tags

vfreex:
Existing `additional-tags` files in distgit repos will be automatically
removed during rebase.

Test build: https://brewweb.engineering.redhat.com/brew/taskinfo?taskID=40266870
Distgit diff: https://pkgs.devel.redhat.com/cgit/containers/openshift-enterprise-cli/commit/?h=rhaos-4.10-rhel-8&id=524587ed73acddaf4e62ffe380c3c968b9f85038

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -54,7 +54,7 @@ CONTAINER_YAML_HEADER = """
 
 # Always ignore these files/folders when rebasing into distgit
 # May be added to based on group/image config
-BASE_IGNORE = [".git", ".oit", "additional-tags"]
+BASE_IGNORE = [".git", ".oit"]
 
 logger = logutil.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


doozerlib/distgit.py
```diff
@@ -515,6 +515,23 @@ def _generate_osbs_image_config(self):
         if arches:
             config_overrides.setdefault('platforms', {})['only'] = arches
 
+        # Request OSBS to apply specified tags to the newly-built image as floating tags.
+        # See https://osbs.readthedocs.io/en/latest/users.html?highlight=tags#image-tags
+        #
+        # Include the UUID in the tags. This will allow other images being rebased
+        # to have a known tag to refer to this image if they depend on it - even
+        # before it is built.
+        floating_tags = {"latest", f"{version}.{self.runtime.uuid}"}
```
sosiouxme, line 524: we did not have `latest` before and while i don't _think_ it will cause conflict in the prod registry, it doesn't seem like a good idea to set it (especially since i keep trying to discourage people from using it at all).

sosiouxme (commented):
otherwise LGTM

doozerlib/distgit.py
vfreex: I added `latest` tag just for keeping compatibility. Now the latest tag is removed.

sosiouxme:
thanks, this looks good.

PR #525, opened 2021-10-12
Pin pygit2 to 1.6

joepvd:
pygit2 1.7 requires libgit 1.3, which is not in the repos yet. Also,
python3.6 support is dropped, which we require on buildvm.

Also, use python3.6 to run the tests in tox to ensure parity in test
environments and production environment.


thiagoalessio (approved):
/lgtm

PR #526, opened 2021-10-21
Permit completeBefore to unbounded during get_latest_build

jupierce:


vfreex:
Tests need to be fixed in this PR, and a rebase.

PR #527, opened 2021-11-01
Prevent leading zero in nvr semver

jupierce:
The semver standard prevents numeric elements of the pre-release portion of a version from starting with `0`. This means that doozer could rarely create invalid semvers like `.git.0123456`. We will change to be `.g0123456` so these fields are always non-numeric.

### DIFF
--- a/doozerlib/assembly.py
+++ b/doozerlib/assembly.py
@@ -9,7 +9,7 @@ from doozerlib.model import Missing, Model, ListModel
 class AssemblyTypes(Enum):
     STREAM = "stream"  # Default assembly type - indicates continuous build and no basis event
     STANDARD = "standard"  # All constraints / checks enforced (e.g. consistent RPMs / siblings)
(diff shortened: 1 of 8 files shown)


yselkowitz:
Wondering if it shouldn't just be `g<short-hash>`, along the lines of `git describe`?

joepvd:
Think this does what says on the tin, yet not comfortable enough for a /lgtm as I have very little idea who or what depends on the current scheme.

Should this first be limited to 4.10 at first and be subject to backports? Or we just roll with it? Should this be documented in all releases with a changelog?

yselkowitz:
Just wondering if there needs to be a test added wrt continuing to recognize the existing `.git.<commit-hash>` form, as this seems to be important as mentioned in the code comments?

vfreex (approved):

vfreex:
No code in Elliott is affected by the new format. But I created this PR https://github.com/openshift/elliott/pull/243 anyway to update examples, comments, and tests to keep consistency.

PR #528, opened 2021-11-01
[ART-3482] Source k8s environment variables for rebase from hyperkube Dockerfile

jupierce:
https://coreos.slack.com/archives/CB48XQ4KZ/p1635786347154300?thread_ts=1635439621.084100&cid=CB48XQ4KZ

### DIFF
--- a/doozerlib/metadata.py
+++ b/doozerlib/metadata.py
@@ -3,8 +3,7 @@ import urllib.parse
 import yaml
 from collections import OrderedDict
 import pathlib
-import json
-import traceback
+import io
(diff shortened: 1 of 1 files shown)


sosiouxme:
/lgtm

concerns that were allayed:
1. Setting these vars goes all the way back to 3.11 when they came out of Godeps.json; all that alternate stuff is being ripped out. But it looks like the Dockerfile is a universal source of truth on this now.
2. These vars got set on other images, like console, but they were probably wrong (based on old tags and such) and not necessary.

PR #536, opened 2021-11-22
[ART-3429] gen-assembly: include upgrades field

vfreex:


doozerlib/cli/release_gen_assembly.py
```diff
@@ -27,12 +30,25 @@ def releases_gen_assembly(ctx, name):
...
+              help="When using --auto-previous, override content from stable channel - primarily for testing")
+@click.option("--graph-content-candidate", metavar='JSON_FILE', required=False,
+              help="When using --auto-previous, override content from candidate channel - primarily for testing")
+@click.option("--suggestions-url", metavar='SUGGESTIONS_URL', required=False,
+              default="https://raw.githubusercontent.com/openshift/cincinnati-graph-data/master/build-suggestions/",
+              help="When using --auto-previous, set custom suggestions URL, load from {major}-{minor}-{arch}.yaml")
 @pass_runtime
 @click.pass_context
-def gen_assembly_from_releases(ctx, runtime, nightlies, standards, custom):
+def gen_assembly_from_releases(ctx, runtime: Runtime, nightlies: Tuple[str, ...], standards: Tuple[str, ...], custom: bool, in_flight: Optional[str], previous_list: Tuple[str, ...], auto_previous: bool, graph_url: Optional[str], graph_content_stable: Optional[str], graph_content_candidate: Optional[str], suggestions_url: Optional[str]):
```
joepvd, line 48: This is a lot of options added to the invocation that in practice will not change. Do we need to have this as variable as this? Maybe we can have them defined in constants.py, so they can be overridden there if we need to experiment during development?

vfreex: I also doubt they will change. I just blindly copied those options and some code from the existing `release:calc-previous` command.

vfreex:
Rebased

joepvd:
The return of the curious test failure :/

doozerlib/cli/release_gen_assembly.py
```diff
@@ -50,6 +66,31 @@ def exit_with_error(msg):
...
+        # gen_assembly_name should be in the form of `fc.0`, `rc.1`, or `4.10.1`
+        if gen_assembly_name.startswith("fc") or gen_assembly_name.startswith("rc"):
+            major_minor = runtime.get_minor_version()  # x.y
+            version = f"{major_minor}.0-{gen_assembly_name}"
+        else:
+            version = gen_assembly_name
+        for arch in runtime.arches:
+            logger.info("Calculating previous list for %s", arch)
+            previous_list = util.get_release_calc_previous(version, arch, graph_url, graph_content_stable, graph_content_candidate, suggestions_url)
+            final_previous_list |= set(map(VersionInfo.parse, previous_list))
```
thegreyd, line 91: Can you add a comment here explaining that we are merging for all arches - since individual arch difference doesn't matter - a note for future developers reading the code

doozerlib/cli/release_gen_assembly.py
```diff
@@ -27,12 +30,25 @@ def releases_gen_assembly(ctx, name):
 @click.option('--standard', 'standards', metavar='4.y.z-ARCH', default=[], multiple=True, help='The name and arch of an official release (e.g. 4.8.3-x86_64) where ARCH in [x86_64, s390x, ppc64le, aarch64].')
 @click.option("--custom", default=False, is_flag=True,
               help="If specified, weaker conformance criteria are applied (e.g. a nightly is not required for every arch).")
+@click.option('--in-flight', 'in_flight', metavar='EDGE', help='An in-flight release that can upgrade to this release')
+@click.option('--previous', 'previous_list', metavar='EDGES', default=[], multiple=True, help='A list of releases that can upgrade to this release')
```
thegreyd, line 34: Is there a need for this flag? if not I'd vote for skipping it

doozerlib/cli/release_gen_assembly.py
thegreyd: I agree with joep, don't see any value in adding them

thegreyd:
Hey @vfreex , this story came out a few days ago, after we messed up an in_flight_ prev https://issues.redhat.com/browse/ART-3705 
Basically the rationale is - right now in_flight_prev is lost in that long list of edges, and it's hard to review. 
So we would like it to be a separate required field - could be any sensible name "in_flight_prev" is probably 
the easiest. Schema wise could be a subkey of "upgrades", don't have a strong preference.

But basically making in_flight_prev explicit would be good for 1) review + 2) promote job validation - check if the assembly definition is the latest/greatest in releases.yaml 

Would appreciate your thoughts.

PR #537, opened 2021-11-22
check ratelimit

Ximinhan:
Our health check job start multi branch at the same time, which maybe cause API call exceed rate limit, check rate limit before do anything if the number of requests remaining in the current rate limit window less than 1000, sleep until next window come.

### DIFF
--- a/doozerlib/cli/images_streams.py
+++ b/doozerlib/cli/images_streams.py
@@ -5,6 +5,7 @@ import yaml
 import json
 import hashlib
 import time
+import datetime
 import random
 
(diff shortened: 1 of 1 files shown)


sosiouxme (approved):
/lgtm
however note I did not test it, I'm assuming you did and it worked fine.
it's cool that github provides these tools for managing rate limits... did not know about these.

Ximinhan:
Unexpectedly, we still encountered limit hopefully it can help
ref:https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting

PR #538, opened 2021-11-22
don't need always pull base image

Ximinhan:


sosiouxme:
i think i like still picking up the latest when it appears :) I did a little cleanup, hopefully it helps with this problem. buildvm-maint is supposed to GC old images, not sure what was happening here to fill up the storage but there's a lot more room now.

PR #539, opened 2021-11-23
container.yml: Add PATCH to tags

joepvd:
v3.11 relies on images being pullable through with a tag like
`:v3.11.345`. This change ensures the patch version is included in
`container.yml`.


vfreex (approved):
/lgtm

PR #544, opened 2021-12-15
Fix submodule clone bug

vfreex:
Fix a bug that prevents submodules from being cloned if cache dir is
disabled.


thiagoalessio (approved):
/lgtm

PR #546, opened 2021-12-16
Enable gen_payload to apply directly and update multi release controller

jupierce:
This change enables gen_payload to generate heterogeneous release payloads
and update the `multi` release controller instance. It also transitions to make
gen_payload directly apply changes to the CI cluster imagestreams. 
The build_sync job should no longer try to run a mirroring operation or update the imagestreams itself.

### DIFF
--- a/doozerlib/assembly_inspector.py
+++ b/doozerlib/assembly_inspector.py
@@ -63,7 +63,7 @@ class AssemblyInspector:
         """
         Analyzes an RHCOS build to check whether the installed packages are consistent with:
         1. package NVRs defined at the group dependency level
-        2. package NVRs defiend at the rhcos dependency level
+        2. package NVRs defined at the rhcos dependency level
         3. package NVRs of any RPMs built in this assembly/group
         :param rhcos_build: The RHCOS build to analyze.
         :return: Returns a (potentially empty) list of inconsistencies in the build.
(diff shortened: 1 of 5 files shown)


sosiouxme:
not sure why https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/art-tools/job/doozer/view/change-requests/job/PR-546/6/console failed to post its comment but there are some legit complaints in there.

jupierce:
@sosiouxme - I think the doozer code is clean now: https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/art-tools/job/doozer/view/change-requests/job/PR-546/9/consoleFull

doozerlib/rhcos.py
```diff
@@ -171,6 +171,9 @@ def __init__(self, runtime, pullspec_or_build_id: str, brew_arch: str):
     def __repr__(self):
         return f'RHCOSBuild:{self.brew_arch}:{self.build_id}'
 
+    def __repr__(self):
```
thegreyd, line 174: @jupierce Was this added in error? Looks duplicate to the one above

jupierce: Yep - looks like a rebase artifact I miss. Removed. Thanks!

joepvd (commented):
Very much like the direction this will be bringing us!

Must admit I have had a hard time reading the proposed change. It'd be awesome if the contents of the method `release_gen_payload` would be split up in other methods. That should probably also help with putting the code under test.

PR #548, opened 2022-01-10
[ART-3665] flatten out settings file extensions iteration to avoid redundant cycles

locriandev:
I've used a list comprehension to derive this list: `['yaml', 'YAML', 'yml', 'YML', 'json', 'JSON']` from the original [VALID_EXT](https://github.com/openshift/doozer/blob/master/doozerlib/dotconfig.py#L11) list. The nested loop can be then flattened out.


joepvd (approved):
/lgtm

We might want to have the same in https://github.com/openshift/elliott/blob/master/elliottlib/dotconfig.py#L60 😬

PR #551, opened 2022-01-21
Allow INCONSISTENT_RHCOS_RPMS in stream

joepvd:
In e.g. [this build-sync job](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-sync/21416/console),
there was no attempt to produce a nightly, because RHCOS had different
rpm contents between the different arches:

```yaml
- code: INCONSISTENT_RHCOS_RPMS
  msg: 'Found RHCOS inconsistencies in builds {False: [RHCOSBuild:x86_64:410.84.202201210604-0, RHCOSBuild:ppc64le:410.84.202201201604-0, RHCOSBuild:s390x:410.84.202201201602-0, RHCOSBuild:aarch64:410.84.202201210203-0], True: []}: {''ignition'': {''ignition-2.13.0-2.rhaos4.10.el8'', ''ignition-2.12.0-3.rhaos4.10.el8''}}'
  permitted: false
```

This is a situation that should be allowed as a stream assembly, as
rhcos can be produced with some delay, while it obviously should be
flagged down when creating a named assembly.

### DIFF
--- a/doozerlib/assembly.py
+++ b/doozerlib/assembly.py
@@ -297,6 +297,10 @@ def assembly_permits(releases_config: Model, assembly: typing.Optional[str]) -> 
             {
                 'code': 'CONFLICTING_GROUP_RPM_INSTALLED',
                 'component': 'rhcos'
+            },
+            {
+                'code': 'INCONSISTENT_RHCOS_RPMS',
+                'component': 'rhcos'
             }
         ]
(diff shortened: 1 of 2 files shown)


joepvd:
Consensus is that this is a liberty we should not take.
/close

PR #552, opened 2022-01-27
[ART-1515] Remove trailing slash from data_path to prevent GitDataBranchException

locriandev:


vfreex (approved):
Technically, the URL with trailing `/` is different from the one without `/`. So that is expected.

But I am fine to merge this.

PR #553, opened 2022-01-28
Permit errors from start-build

jupierce:
In some situations, start-build returns an error immediately
if images don't exist. The intention of this flow was to
try the build and eventually things will reconcile. It is ok
for images to be missing on the first runs of reconciliation.


joepvd (approved):
/lgtm

PR #556, opened 2022-02-11
Delete payload istags when they are no longer built

jupierce:
The previous implementation removed old imagestream tags
from the imagestream monitored by the release controller.
However, the release controller considers tags in the
status field. Removing tags from the status field
requires an explicit deletion operation.

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -1,6 +1,6 @@
 import datetime
 import hashlib
-import pathlib
+import traceback
 import sys
 import json
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


joepvd:
Nice! Think this could use a safety check that the number of items to prune does not exceed a very small number.

jupierce:
@joepvd - we backup these imagestreams on each build-sync, so there is a means to recover if things go wrong. If I do add an error, that implies to me that we have another command line argument and another jenkins job parameter to override that error -- and that is getting into a fairly esoteric space.
Given that, I think relying on the backups is our best option to handle the many 'what-if' scenarios.

vfreex (approved):
/lgtm

PR #564, opened 2022-03-02
images:rebase: Sort tags and content sets

vfreex:
It's often not very obvious to see diffs on distgit. Sorting tags and content sets will create a smaller diff in distgit.


joepvd:
/lgtm

oh yes! Been wanting this so long! Thanks for scratching this itch :)

PR #566, opened 2022-03-03
Always use Cachito remote source name cachito-gomod-with-deps

vfreex:
Existing repos has hardcoded this name (e.g.https://github.com/openshift/etcd/blob/f99cada055fe83dc3cefb546dea95547d7da727d/Dockerfile.art#L4).

For backward compatibility, let's always use `cachito-gomod-with-deps` even if gomod is not used.


thiagoalessio (approved):
/lgtm

PR #573, opened 2022-03-11
Allow an RPM is an image dependency to not appear in the final image

jupierce:
This actually ocurred during [ART-3875](https://issues.redhat.com/browse/ART-3875) where the ironic image
installed grub2 in a stage, but it was not included in the final
stage.


vfreex (approved):
/lgtm

PR #575, opened 2022-03-21
Make retries image builds configurable

joepvd:
In the past, we had nightly full product builds. Then it made sense to
retry a failed image build.

Currently, our automation detects that a build is needed, and will spin
off a new build when necessary.

In the case of an image that takes a lot of timne to build, say
installer-artifacts, it takes around 1.5 hours for a single build. This
means that with three failing retries, product builds are halted for
three or four hours.

With this change, `doozer images:build` will grow a command line option
`--build-retries` that defaults to `1`, so that, in the default case:
- retries will continue to happen through other means
- images that need rebuilding will get rebuilt faster

For builds on a schedule, like OCP3.11, it makes sense to set retries to
3.

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -874,8 +874,9 @@ def print_build_metrics(runtime):
 @click.option("--filter-by-os", default=None, metavar="ARCH",
               help="Specify an exact arch to push (golang name e.g. 'amd64').")
 @click.option('--dry-run', default=False, is_flag=True, help='Do not build anything, but only print build operations.')
+@click.option('--build-retries', type=int, default=1, help='Number of build attempts for an osbs build')
 @pass_runtime
-def images_build_image(runtime, repo_type, repo, push_to_defaults, push_to, scratch, threads, filter_by_os, dry_run):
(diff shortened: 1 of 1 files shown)


vfreex:
Retrying here might be useful for 3.11 build. Also would like to hear thoughts from other team members.

vfreex (approved):
/lgtm

doozerlib/cli/__main__.py
```diff
@@ -874,8 +874,9 @@ def print_build_metrics(runtime):
 @click.option("--filter-by-os", default=None, metavar="ARCH",
               help="Specify an exact arch to push (golang name e.g. 'amd64').")
 @click.option('--dry-run', default=False, is_flag=True, help='Do not build anything, but only print build operations.')
+@click.option('--build-retries', default=1, help='Number of retries for an osbs build')
```
vfreex, line 877: @joepvd I believe `type=int` is needed here, otherwise we will get TypeError.

vfreex (changes requested):
Only 1 issue. Otherwise LGTM.

doozerlib/cli/__main__.py
joepvd: Thanks! I think this is fixed in the new iteration

PR #576, opened 2022-03-22
get_latest_build: permit unbounded brew event

vfreex:
Based on https://github.com/openshift/doozer/pull/526.

If `complete_before_event < 0`, `get_latest_build` will return the latest
build with no constraint on brew event. This should fix the tag push issue
in 3.11.

This PR also consolidates the implementation in `doozerlib` and
`elliottlib`: Include tag names of the build in `_tags` field.


joepvd (approved):
/lgtm

PR #580, opened 2022-03-29
Remove rpm-py-installer

vfreex:
This package is RHEL-only, which also caused a lot of build failures in the past.
It is used to compare rpm versions. I would like to port it into pure C
implementation.

### DIFF
--- a/tests/test_rpm_utils.py
+++ b/tests/test_rpm_utils.py
@@ -1,0 +1,117 @@
+import imp
+from unittest import TestCase
+from doozerlib import rpm_utils
+
+
+class TestRPMUtils(TestCase):
+    def test_rpmverCmp(self):
+        # Tests are extracted from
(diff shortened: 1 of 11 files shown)


tests/test_rpm_utils.py
```diff
@@ -0,0 +1,117 @@
+import imp
+from unittest import TestCase
+from doozerlib import rpm_utils
+
+
+class TestRPMUtils(TestCase):
+    def test_rpmverCmp(self):
+        # Tests are extracted from
+        # https://github.com/rpm-software-management/rpm/blob/9e4caf0fc536d1244b298abd9dc4c535b6560d69/tests/rpmvercmp.at#L1
```
joepvd, line 9: 😍

joepvd:
Maybe a comment with a pointer to the elliott code is helpful?

/lgtm

PR #581, opened 2022-03-30
[ART-3674] embed publish arg into gen-payload

Ximinhan:
In[ build-sync job](https://github.com/openshift/aos-cd-jobs/blob/master/jobs/build/build-sync/build.groovy#L76-L99) there is a publish arg used to push IS into registry.ci, the implement in jenkins is read the out put then reload, I want to embed it into doozer gen-payload

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -449,11 +449,6 @@ read and propagate/expose this annotation in its display of the release image.
                 with output_path.joinpath(f"updated-tags-for.{imagestream_namespace}.{imagestream_name}{'-partial' if incomplete_payload_update else ''}.yaml").open("w+", encoding="utf-8") as out_file:
                     istream_spec = PayloadGenerator.build_payload_imagestream(imagestream_name, imagestream_namespace, istags, assembly_issues)
                     yaml.safe_dump(istream_spec, out_file, indent=2, default_flow_style=False)
-                    if publish:
-                        publish_name = f"{runtime.get_minor_version()}.0-{runtime.assembly}"
-                        publish_reponame = imagestream_namespace.replace("ocp", "release")
-                        publish_image = f"registry.ci.openshift.org/{imagestream_namespace}/{publish_reponame}:{publish_name}"
-                        exectools.cmd_assert(f'oc adm release new --to-image={publish_image} --name {publish_name} --reference-mode=source -n {imagestream_namespace} --from-image-stream {imagestream_name}', retries=3)
 
             if apply:
                 with oc.project(imagestream_namespace):
@@ -510,6 +505,13 @@ read and propagate/expose this annotation in its display of the release image.
 
                     if adding_tags:
                         logger.warning(f'The following tag names are net new to {imagestream_namespace}:{imagestream_name}: {adding_tags}')
+
+                if publish:
+                    arch_suffix = go_suffix_for_arch(arch)
+                    priv_suffix = "-priv" if private_mode else ""
+                    publish_name = f"{runtime.get_minor_version()}.0-{runtime.assembly}"
+                    publish_image = f"registry.ci.openshift.org/{imagestream_namespace}/release{arch_suffix}{priv_suffix}:{publish_name}"
+                    exectools.cmd_assert(f'oc adm release new --to-image={publish_image} --name {publish_name} --reference-mode=source -n {imagestream_namespace} --from-image-stream {imagestream_name}', retries=3)
 
     # We now generate the artifacts to create heterogeneous release payloads. A heterogeneous or 'multi' release
     # payload is a manifest list (i.e. it consists of N release payload manifests, one for each arch). The release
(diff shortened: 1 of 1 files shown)


doozerlib/cli/release_gen_payload.py
```diff
@@ -447,6 +449,11 @@ def release_gen_payload(runtime: Runtime, is_name: str, is_namespace: str, organ
                 with output_path.joinpath(f"updated-tags-for.{imagestream_namespace}.{imagestream_name}{'-partial' if incomplete_payload_update else ''}.yaml").open("w+", encoding="utf-8") as out_file:
                     istream_spec = PayloadGenerator.build_payload_imagestream(imagestream_name, imagestream_namespace, istags, assembly_issues)
                     yaml.safe_dump(istream_spec, out_file, indent=2, default_flow_style=False)
+                    if publish:
+                        publish_name = f"{runtime.get_minor_version()}.0-{runtime.assembly}"
+                        publish_reponame = imagestream_namespace.replace("ocp", "release")
+                        publish_image = f"registry.ci.openshift.org/{imagestream_namespace}/{publish_reponame}:{publish_name}"
+                        exectools.cmd_assert(f'oc adm release new --to-image={publish_image} --name {publish_name} --reference-mode=source -n {imagestream_namespace} --from-image-stream {imagestream_name}', retries=3)
```
vfreex, line 456: This is too early. When this is executed, the image stream doesn't exist yet.

doozerlib/cli/release_gen_payload.py
```diff
@@ -447,6 +449,11 @@ def release_gen_payload(runtime: Runtime, is_name: str, is_namespace: str, organ
                 with output_path.joinpath(f"updated-tags-for.{imagestream_namespace}.{imagestream_name}{'-partial' if incomplete_payload_update else ''}.yaml").open("w+", encoding="utf-8") as out_file:
                     istream_spec = PayloadGenerator.build_payload_imagestream(imagestream_name, imagestream_namespace, istags, assembly_issues)
                     yaml.safe_dump(istream_spec, out_file, indent=2, default_flow_style=False)
+                    if publish:
+                        publish_name = f"{runtime.get_minor_version()}.0-{runtime.assembly}"
+                        publish_reponame = imagestream_namespace.replace("ocp", "release")
```
vfreex, line 454: This seems to be dangerous to do a string replacement because you are making an assumption that there is one but only one `ocp` substring in the imagestream_namespace. If it changes, this will probably overwrite things unexpectedly. I believe to be explicitly configured either by command line arguments or config files.

vfreex (changes requested):

doozerlib/cli/release_gen_payload.py
```diff
@@ -504,6 +506,16 @@ def update_single_arch_istags(apiobj: oc.APIObject):
                     if adding_tags:
                         logger.warning(f'The following tag names are net new to {imagestream_namespace}:{imagestream_name}: {adding_tags}')
 
+            if publish:
+                arch_suffix = go_suffix_for_arch(arch)
+                priv_suffix = "-priv" if private_mode else ""
+                imagestream_name = f"{base_imagestream_name}{arch_suffix}{priv_suffix}"
+                imagestream_namespace = f"release{arch_suffix}{priv_suffix}"
+                publish_name = f"{runtime.get_minor_version()}.0-{runtime.assembly}"
+                publish_reponame = imagestream_namespace.replace("ocp", "release")
```
vfreex, line 515: This seems to be dangerous to do a string replacement because you are making an assumption that there is one but only one ocp substring in the imagestream_namespace. If a different namespace is given by the parameter, this will probably overwrite things unexpectedly. I believe to be explicitly configured either by command line arguments or config files.

doozerlib/cli/release_gen_payload.py
```diff
@@ -504,6 +506,16 @@ def update_single_arch_istags(apiobj: oc.APIObject):
                     if adding_tags:
                         logger.warning(f'The following tag names are net new to {imagestream_namespace}:{imagestream_name}: {adding_tags}')
 
+            if publish:
```
vfreex, line 509: If `--apply` is not used this will cause an error because the referenced image name doesn't exist.

doozerlib/cli/release_gen_payload.py
```diff
@@ -114,11 +114,13 @@ def modify_and_replace_api_object(api_obj: oc.APIObject, modifier_func: Callable
               help="If doozer should also create a release payload for multi-arch/heterogeneous clusters.")
 @click.option("--moist-run", default=False, is_flag=True,
               help="Performing mirroring/etc but to not actually update imagestreams.")
+@click.option("--publish", default=False, is_flag=True,
```
vfreex, line 117: Parameter name `--publish` sounds misleading.

vfreex (changes requested):

doozerlib/cli/release_gen_payload.py
Ximinhan: this parameter is inherit from jenkins job's parameter

vfreex: But this is not a jenkins job. Jenkins job assumes some certain use cases but our Doozer may need more flexibility. 
An option name like `--also-publish-to=namespace` or split the publishing step into separate Doozer verb / artcd step could be a better idea.

sosiouxme: realistically i don't think there's a use for this doozer verb outside a job like build-sync. so inlining job logic makes sense.

what's misleading about `--publish`? it creates a release image where people can get to it to test it. registry.ci isn't exactly public but otherwise... that's publishing :) however the param could probably benefit from a more explicit name, to distinguish from what the `promote` job does. maybe `--publish-test-image`?

sosiouxme:
after the rebase, this will look something like https://github.com/openshift/doozer/pull/673

PR #582, opened 2022-03-31
[ART-3921] Images using private builders should also be private

vfreex:
There is a race condition. An image should wait for its builder image to be rebased to determine if there's a private fix in the builder.


joepvd (approved):
/lgtm

PR #584, opened 2022-04-04
inspect:stream: Add strict mode

joepvd:
Currently, ocp4_scan relies on the exit code of inspect:stream to
determine if an rhcos build needs to be kicked off. As rhcos has been
failing, we have temporarily allowed build-sync to ignore inconsistent
rpms in rhcos to produce nightlies. This circumstance currently leads to
not triggering nightly builds.

This commit adds a `--strict` command line option, defaulting to
`False`. The current behavior is kept, but with invoking it with this
new flag, ocp4_scan will trigger new rhcos builds regardless of whether
the inconsistency is allowed.


thiagoalessio (approved):
/lgtm

NOICE !!!

PR #586, opened 2022-04-19
Improve heterogeneous release payload assembly performance

jupierce:
By reusing manifest-lists created by brew, this change
dramatically cuts down on the manifest lists that must be
stitched together -- significantly reducing the time to
construct the heterogeneous payload.


thiagoalessio (approved):
/lgtm

PR #587, opened 2022-04-22
scan_sources: Deal with http error in cgit_atom_feed

joepvd:
Catch the exception so the right error will be logged in the calling
function.


vfreex (changes requested):
That retry call is required to wait for cgit cache. Every time we push a commit to distgit, it may take a few minutes to show up on cgit.

vfreex:
ok, looked at urlopen_assert, but looks like the exception should be RetryException on max retries attempts.
We should reraise the exception with a descriptive message rather than ignore it and return None.

PR #590, opened 2022-04-27
When runtime initializes, check if assembly is defined in releases.yml

locriandev:


vfreex (changes requested):
The current behavior is expected and documented. We need the ability to create builds without polluting our nightlies and other releases. 

If you have a good reason to change the behavior, talk with @jupierce :)

locriandev:
closed after [discussing with the team](https://coreos.slack.com/archives/GDBRP5YJH/p1651045866999139)

PR #593, opened 2022-05-17
Support specifying `packages` option for Cachito

vfreex:
Allow user to customize `packages` option for Cachito configuration.
See https://osbs.readthedocs.io/en/osbs_ocp3/users.html#remote-source-keys for details.


vfreex:
https://github.com/openshift/ocp-build-data-validator/pull/65/ is the schema.

PR #596, opened 2022-05-25
[ART-4070] Add valid-subscription annotation to bundle

thiagoalessio:
`operators.openshift.io/valid-subscription` is now present in `annotations.yml`
and in the bundle's Dockerfile (as a `LABEL`)


vfreex:
Is it a required label?

thiagoalessio:
eeeeeerrr ... cherry-pick comments on the wrong PR. they should go to ocp-build-data, not doozer

PR #607, opened 2022-06-20
Pin click to 8.0.*

joepvd:
Click 8.1 does not support python3.6, which is what we have in
production environment.


thiagoalessio (approved):
/lgtm

PR #608, opened 2022-06-20
Set https_proxy env var for covscan image builds

locriandev:
https://coreos.slack.com/archives/GDBRP5YJH/p1655720187867989

### DIFF
--- a/doozerlib/coverity.py
+++ b/doozerlib/coverity.py
@@ -273,6 +273,9 @@ FROM {parent_image_url}
 LABEL DOOZER_COVSCAN_GROUP_PARENT={cc.runtime.group_config.name}
 USER 0
 
+# Set https proxy
+ENV https_proxy http://proxy.util.phx1.redhat.com:8080
+
 # Add typical build repos to the image, but don't add to /etc/yum.repos.d
 # until we know whether we are on el7 or el8. As of 4.8, repos are only
 # appropriate for el8, so this repo file should only be installed in el8.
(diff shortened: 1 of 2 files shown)


doozerlib/coverity.py
```diff
@@ -273,6 +273,9 @@ def _covscan_prepare_parent(cc: CoverityContext, parent_image_name, parent_tag)
 LABEL DOOZER_COVSCAN_GROUP_PARENT={cc.runtime.group_config.name}
 USER 0
 
+# Set https proxy
+ENV https_proxy http://proxy.util.phx1.redhat.com:8080
```
joepvd, line 277: Prefer not to hardcode a string in this location. Maybe query `os.environ` for the existence of `(https?|no)_proxy` vars and set them if defined?

vfreex: We may change to another proxy over time. Agree this shouldn't been hardcoded.

An option could be using `ARG` directive in Dockerfile then pass proxy related environment variables to the podman command line.

vfreex: We may change to another proxy over time. Agree this shouldn't been hardcoded.

An option could be using `ARG` directive in Dockerfile then pass proxy related environment variables to the podman command line.

vfreex (changes requested):

locriandev:
If merged, will require https://github.com/openshift/aos-cd-jobs/pull/3113

vfreex (approved):
/lgtm

PR #613, opened 2022-06-24
images_streams: enable rhel-9 transforms

sosiouxme:


joepvd:
/lgtm

https://github.com/openshift/doozer/pull/612 was also up, don't care which one goes

sosiouxme:
let's start with the minimal change here :)

PR #618, opened 2022-07-05
[ART-3883] multiple rhcos containers

sosiouxme:
https://issues.redhat.com/browse/ART-3883

We will need to configurably inject multiple containers from an RHCOS build into the payload. Eventually `machine-os-content` may not be the reference container and may not be present at all. https://github.com/openshift/ocp-build-data/pull/1773/files shows how we might expect config to look for the first additional container. https://github.com/openshift/ocp-build-data-validator/pull/78 enables `rhel-coreos-8` in assembly definitions.

If the changes seem a bit much to review, it will likely be easier to review one commit at a time, as each represents a discrete change.

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -943,11 +943,12 @@ class PayloadGenerator:
                 final_members[tag_name] = pod_entry
 
         rhcos_build: RHCOSBuildInspector = assembly_inspector.get_rhcos_build(arch)
-        final_members['machine-os-content'] = PayloadGenerator.PayloadEntry(
-            dest_pullspec=rhcos_build.get_image_pullspec(),
-            rhcos_build=rhcos_build,
-            issues=list(),
-        )
+        for container_config in rhcos_build.get_container_configs():
+            final_members[container_config.name] = PayloadGenerator.PayloadEntry(
+                dest_pullspec=rhcos_build.get_container_pullspec(container_config),
+                rhcos_build=rhcos_build,
+                issues=list(),
+            )
 
         # Final members should have all tags populated.
         return final_members
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -12,7 +12,7 @@ import openshift as oc
 from doozerlib.rpm_utils import parse_nvr
 
 from doozerlib.brew import KojiWrapper
-from doozerlib.rhcos import RHCOSBuildInspector
+from doozerlib.rhcos import RHCOSBuildInspector, RhcosMissingContainerException
 from doozerlib.cli import cli, pass_runtime
 from doozerlib.image import ImageMetadata, BrewBuildImageInspector, ArchiveImageInspector
 from doozerlib.assembly_inspector import AssemblyInspector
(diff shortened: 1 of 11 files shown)


doozerlib/rhcos.py
```diff
@@ -9,18 +9,73 @@
...
 from doozerlib import brew
 
 RHCOS_BASE_URL = "https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases"
+default_primary_container = dict(  # historically the only RHCOS container
+    name="machine-os-content",
+    build_metadata_key="oscontainer",
+    primary=True)
 
 
-def rhcos_content_tag(runtime) -> str:
```
sosiouxme, line 18: turns out nothing references this; we may need to resurrect it somehow when RHCOS is building for rhel8 and rhel9.

doozerlib/cli/release_gen_payload.py
```diff
@@ -894,8 +897,8 @@ def get_mirroring_destination(sha256: str, dest_repo: str) -> str:
         tag = sha256.replace(":", "-")  # sha256:abcdef -> sha256-abcdef
         return f"{dest_repo}:{tag}"
 
-    @staticmethod
-    def find_payload_entries(assembly_inspector: AssemblyInspector, arch: str, dest_repo: str) -> Dict[str, PayloadEntry]:
+    @classmethod
+    def find_payload_entries(clazz, assembly_inspector: AssemblyInspector, arch: str, dest_repo: str) -> (Dict[str, PayloadEntry], List[AssemblyIssue]):
```
sosiouxme, line 901: i broke this method up because it was impossible to unit test. i only ended up unit testing the new stuff but more tests are welcome :)

doozerlib/cli/scan_sources.py
```diff
@@ -265,30 +265,30 @@ def _detect_rhcos_status(runtime, kubeconfig) -> list:
     for arch in runtime.arches:
         for private in (False, True):
             name = f"{version}-{arch}{'-priv' if private else ''}"
-            tagged_mosc_id = _tagged_mosc_id(kubeconfig, version, arch, private)
+            tagged_rhcos_id = _tagged_rhcos_id(kubeconfig, rhcos.get_primary_container_name(runtime), version, arch, private)
```
vfreex, line 268: Do we need to check other tags? Are they guaranteed to be updated at the same time?

vfreex (commented):
Generally LGTM. Just have some nits and questions.

doozerlib/rhcos.py
```diff
@@ -9,18 +9,85 @@
...
+
+def get_primary_container_conf(runtime):
+    """
+    look up the group.yml-configured primary RHCOS container for this group.
+    @return Model with entries for name and build_metadata_key
+    """
+    for tag in get_container_configs(runtime):
+        if tag.primary:
+            return tag
+    raise Exception("Need to provide a group.yml rhcos.payload_tags entry with primary=true")
```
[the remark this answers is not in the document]
sosiouxme: if i care at all what type of exception is raised, i'll generally define one specifically for it, so i can test for it. i don't think i care here...

sosiouxme (commented):
@vfreex thanks! i also found another minor bug to fix while testing suggested changes... will update

vfreex (approved):
/lgtm

PR #622, opened 2022-07-18
get-nightlies: revamp for deeper comparison

sosiouxme:
Comparing timestamps on nightlies can only go so far.
For a more reliable indicator that all the nightlies selected are "the same thing" get-nightlies now compares contents quite deeply.
This should be reliable enough to feed into gen-assembly or to use in jobs checking that nightlies are available for upcoming releases.

It also allows specifying a nightly you'd like in the set and other improvements.

```
$ time doozer -q -g openshift-4.10 get-nightlies
4.10.0-0.nightly-ppc64le-2022-07-18-094300 4.10.0-0.nightly-s390x-2022-07-16-173211 4.10.0-0.nightly-arm64-2022-07-16-173300 4.10.0-0.nightly-2022-07-18-094217

real	0m30.512s
user	1m7.026s
sys	0m16.111s

$ time doozer -q -g openshift-4.10 get-nightlies --matching 4.10.0-0.nightly-2022-07-18-094217
4.10.0-0.nightly-ppc64le-2022-07-18-094300 4.10.0-0.nightly-s390x-2022-07-16-173211 4.10.0-0.nightly-arm64-2022-07-16-173300 4.10.0-0.nightly-2022-07-18-094217

real	0m16.225s
user	1m3.965s
sys	0m15.145s

```

### DIFF
--- a/doozerlib/cli/get_nightlies.py
+++ b/doozerlib/cli/get_nightlies.py
@@ -88,30 +88,33 @@ async def get_nightlies(runtime, matching: List[str], exclude_arch: List[str], a
 
     # make lists of nightly objects per arch
     try:
-        nightlies_for_arch: Dict[str, List[Dict]] = find_rc_nightlies(runtime, include_arches, allow_pending, allow_rejected, matching)
+        nightlies_for_arch: Dict[str, List[Nightly]] = {
+            arch: [Nightly(nightly_info=n) for n in nightlies]
+            for arch, nightlies in find_rc_nightlies(runtime, include_arches, allow_pending, allow_rejected, matching).items()
+        }
     except NoMatchingNightlyException as ex:
         util.red_print(ex)
         exit(1)
 
     # retrieve release info for each nightly image (with concurrency)
     await asyncio.gather(*[
-        populate_nightly_release_data(nightly)
+        nightly.populate_nightly_release_data()
         for arch, nightlies in nightlies_for_arch.items()
         for nightly in nightlies
     ])
 
     # find sets of nightlies where all arches have equivalent content
-    eq_sets = []
-    for eq_set in generate_equivalence_sets(nightlies_for_arch):
+    nightly_sets = []
+    for nightly_set in generate_nightly_sets(nightlies_for_arch):
         # check for deeper equivalence
-        await eq_set.populate_deeper_equivalence(runtime)
-        if eq_set.deeper_equivalence():
-            eq_sets.append(eq_set)
-            util.green_print(eq_set.details() if details else eq_set)
-            if len(eq_sets) >= limit:
-                break  # don't spend time looking for more than were requested
+        await nightly_set.populate_nightly_content(runtime)
+        if nightly_set.deeper_equivalence():
+            nightly_sets.append(nightly_set)
+            util.green_print(nightly_set.details() if details else nightly_set)
+            if len(nightly_sets) >= limit:
+                break  # don't spend time checking more than were requested
 
-    if not eq_sets:
+    if not nightly_sets:
         util.red_print("No sets of equivalent nightlies found for given parameters.")
         exit(1)
 
@@ -152,10 +155,10 @@ class EmptyArchException(Exception):
 
 def find_rc_nightlies(runtime, arches: Set[str], allow_pending: bool, allow_rejected: bool, matching: Optional[List[str]] = []) -> Dict[str, List[Dict]]:
     """
-    Find all current nightlies for each arch, in order RC gives them (most
-    recent to oldest). Filter to Accepted unless allow_rejected is true.
-    ref.  https://amd64.ocp.releases.ci.openshift.org/api/v1/releasestream/4.12.0-0.nightly/tags
-    Each nightly looks like:
+    Retrieve current nightly dicts for each arch, in order RC gives them (most
+    recent to oldest). Filter to Accepted unless allow_pending/rejected is true.
+    ref. https://amd64.ocp.releases.ci.openshift.org/api/v1/releasestream/4.12.0-0.nightly/tags
+    Each nightly dict looks like:
     {
       "name": "4.12.0-0.nightly-2022-07-15-132344",
       "phase": "Ready",
@@ -210,7 +213,7 @@ def rc_api_url(tag: str, arch: str) -> str:
     """
     base url for a release tag in release controller.
 
-    @param tag  The RC release tag as a string (e.g. "4.9.0-0.nightly")
+    @param tag  The RC release stream as a string (e.g. "4.9.0-0.nightly")
     @param arch  architecture we are interested in (e.g. "s390x")
     @return e.g. "https://s390x.ocp.releases.ci.openshift.org/api/v1/releasestream/4.9.0-0.nightly-s390x"
     """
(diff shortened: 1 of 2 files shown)


doozerlib/cli/get_nightlies.py
```diff
@@ -1,97 +1,186 @@
...
+from doozerlib.rhcos import RHCOSBuildInspector
 
+logger = logutil.getLogger(__name__)
 
-@cli.command("get-nightlies", short_help="Get sets of Accepted nightlies. A set contains nightly for each arch, "
-                                         "determined by closest timestamps")
-@click.option("--limit", default=1, metavar='NUM', help="Number of sets of passing nightlies to print")
+
+@cli.command("get-nightlies", short_help="Determine set(s) of accepted nightlies with matching contents for all architectures")
+@click.option("--matching", metavar="NIGHTLY_NAME", multiple=True, help="Only report nightlies with the same content as named nightly")
```
thegreyd, line 17: I think the majority usecase for this would be using it with a passing x64 nightly.. any reason we're accepting multiple?

doozerlib/cli/get_nightlies.py
```diff
@@ -1,97 +1,186 @@
...
+@click.option("--allow-pending", is_flag=True, help="Include nightlies that have not completed tests")
+@click.option("--allow-rejected", is_flag=True, help="Include nightlies that have failed tests")
+@click.option("--exclude-arch", metavar="ARCH", multiple=True, help="Exclude arch(es) normally included in this version (multi,aarch64,...)")
+@click.option("--limit", default=1, metavar='NUM', help="Number of sets of nightlies to print")
 @click.option("--rhcos", is_flag=True, help="Print rhcos build id with nightlies")
```
thegreyd, line 22: "doozer --quiet -g openshift-4.10 get-nightlies --rhcos" is not printing the build ids right now..

doozerlib/cli/get_nightlies.py
```diff
@@ -1,97 +1,186 @@
...
+    If results do not include a nightly that you expect to see, check the
+    doozer debug.log where details about equivalence failures are logged.
+    Matching is performed in two phases:
+
+    * The first uses only info from the nightly release images for quick
+      comparison in order to construct candidate equivalent sets of nightlies.
 
+    * The second retrieves image info for all payload content in order to
+      compare group image NVRs and RHCOS RPM content.
+    """
```
thegreyd, line 53: I would add some example invocations here, the normal usecase and the --matching one would help a new user

doozerlib/cli/get_nightlies.py
thegreyd: At first I thought this was accepting nightlies of different arches, `--matching x64_nightly --matching p8_nightly` , but this seems to be intended for `--matching x64_nightly --matching another_x64_nightly`. But I'm not sure if we're restricting the first usecase? A bit confusing

doozerlib/cli/get_nightlies.py
```diff
@@ -107,16 +196,269 @@ def rc_api_url(tag, arch):
...
+    Class to represent a set of nightlies from each arch that are equivalent
+    according to their EquivalenceBase. They can be sorted by creation date
+    (latest timestamp of nightlies in the set)
+    """
+
+    def __init__(self, nightly_for_arch: Dict[str, Dict]):
+        self.nightly_for_arch = nightly_for_arch
+        self.timestamp = max(nightly["releaseInfo"]["config"]["created"] for nightly in nightly_for_arch.values())
+
+    def augment(self, arch: str, nightly: Dict):
```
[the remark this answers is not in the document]
sosiouxme: perhaps better to internalize the intended usage as:
```
    def generate_equivalents_with(self, arch: str, nightly: List[Dict]) -> List[EquivalentSet]:
```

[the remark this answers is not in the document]
sosiouxme: I added some commits and implemented it this way... see if you like it better.

doozerlib/cli/get_nightlies.py
sosiouxme: i tried making it into `--details` instead

doozerlib/cli/get_nightlies.py
thegreyd: :+1: Yeah I don't mind having it in, it's pretty nifty!

doozerlib/cli/get_nightlies.py
thegreyd: thanks for adding this

doozerlib/cli/get_nightlies.py
thegreyd: Tried testing it, but I'm getting too many open files error 
```
Traceback (most recent call last):
  File "/home/sid/.local/bin/doozer", line 11, in <module>
    load_entry_point('rh-doozer', 'console_scripts', 'doozer')()
  File "/home/sid/Projects/art/doozer/doozerlib/cli/__main__.py", line 2234, in main
    cli(obj={})
  File "/usr/lib/python3.10/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.10/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.10/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/click/decorators.py", line 38, in new_func
    return f(get_current_context().obj, *args, **kwargs)
  File "/home/sid/Projects/art/doozer/doozerlib/cli/__init__.py", line 167, in wrapper
    return loop.run_until_complete(f(*args, **kwargs))
  File "/usr/lib64/python3.10/asyncio/base_events.py", line 646, in run_until_complete
    return future.result()
  File "/home/sid/Projects/art/doozer/doozerlib/cli/get_nightlies.py", line 107, in get_nightlies
    await eq_set.populate_deeper_equivalence(runtime)
  File "/home/sid/Projects/art/doozer/doozerlib/cli/get_nightlies.py", line 453, in populate_deeper_equivalence
    await asyncio.gather(*(
  File "/home/sid/Projects/art/doozer/doozerlib/cli/get_nightlies.py", line 343, in populate_deeper_equivalence
    await asyncio.gather(*(
  File "/home/sid/Projects/art/doozer/doozerlib/cli/get_nightlies.py", line 334, in retrieve_image_info_async
    image_json_str, _ = await exectools.cmd_assert_async(
  File "/home/sid/Projects/art/doozer/doozerlib/exectools.py", line 374, in cmd_assert_async
    result, out, err = await cmd_gather_async(cmd_list, text_mode=text_mode, cwd=cwd, set_env=set_env, strip=strip, log_stdout=log_stdout, log_stderr=log_stderr)
  File "/home/sid/Projects/art/doozer/doozerlib/exectools.py", line 309, in cmd_gather_async
    proc = await asyncio.create_subprocess_exec(
  File "/usr/lib64/python3.10/asyncio/subprocess.py", line 218, in create_subprocess_exec
    transport, protocol = await loop.subprocess_exec(
  File "/usr/lib64/python3.10/asyncio/base_events.py", line 1667, in subprocess_exec
    transport = await self._make_subprocess_transport(
  File "/usr/lib64/python3.10/asyncio/unix_events.py", line 207, in _make_subprocess_transport
    transp = _UnixSubprocessTransport(self, protocol, args, shell,
  File "/usr/lib64/python3.10/asyncio/base_subprocess.py", line 36, in __init__
    self._start(args=args, shell=shell, stdin=stdin, stdout=stdout,
  File "/usr/lib64/python3.10/asyncio/unix_events.py", line 799, in _start
    self._proc = subprocess.Popen(
  File "/usr/lib64/python3.10/subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib64/python3.10/subprocess.py", line 1744, in _execute_child
    errpipe_read, errpipe_write = os.pipe()
OSError: [Errno 24] Too many open files
```

thegreyd: Huh so that ^ only happens when I run the command in "Tilix" but not in vscode/pycharm terminal. The output looks good otherwise

doozerlib/cli/get_nightlies.py
```diff
@@ -107,16 +219,277 @@ def rc_api_url(tag, arch):
...
+        """
+        if arch in self.nightly_for_arch:
+            raise EqSetDuplicateArchException(f"Cannot add arch {arch} when already present in {self}")
+
+        new_sets: List[EquivalenceSet] = []
+        for nightly in nightlies:
+            nightly_for_arch = {arch: nightly}  # initialize new set with candidate nightly
+            for exarch, existing in self.nightly_for_arch.items():
+                logger.debug(f"comparing {nightly['name']} and {existing['name']}")
+                if nightly["equivalence"] != existing["equivalence"]:
```
thegreyd, line 433: This block looks inverted to me.. any reason why we can't break when we find the first match, like
```python
if nightly["equivalence"] == existing["equivalence"]:
    nightly_for_arch[exarch] = existing
    break
```

thegreyd: I'm not sure why we're not returning here when the first matching nightly is found for a set.. are we expecting a case where multiple identical nightlies for an arch could match?

doozerlib/cli/get_nightlies.py
```diff
@@ -107,16 +219,277 @@ def rc_api_url(tag, arch):
...
+            for other in nightlies:
+                logger.debug(f"comparing {this['name']} and {other['name']}")
+                if not this["equivalence"].deeper_equivalence(other["equivalence"]):
+                    logger.debug(f"deeper equivalence failed for {self}")
+                    return False
+
+        return True
+
+
+def generate_equivalence_sets(nightlies_for_arch: Dict[str, List[Dict]]) -> List[EquivalenceSet]:
```
thegreyd, line 472: Think this should accept limit, we don't need to process all sets if we're asking for the first accepted one

doozerlib/cli/get_nightlies.py
sosiouxme: interesting. i never saw that, but i have:
```
$ ulimit -n
524288
```
... so it would be hard to run out of open files. i wonder if there's a way to limit the concurrency though... this is the point where it's running `oc image info` against all the images in all the nightlies at once...

doozerlib/cli/get_nightlies.py
```diff
@@ -1,100 +1,212 @@
...
-def rc_api_url(tag, arch):
+    # find sets of nightlies where all arches have equivalent content
+    eq_sets = []
+    for eq_set in generate_equivalence_sets(nightlies_for_arch):
+        # check for deeper equivalence
+        await eq_set.populate_deeper_equivalence(runtime)
+        if eq_set.deeper_equivalence():
+            eq_sets.append(eq_set)
+            util.green_print(eq_set.details() if details else eq_set)
+            if len(eq_sets) >= limit:
```
thegreyd, line 111: Ah I see the limit here and the early break. I see the point of generating all sets since if 1 set is not deeply equivalent we won't have any to return.. hmm can we have deeper_equivalence comparison inside generate_equivalence_sets and return early?

doozerlib/cli/get_nightlies.py
```diff
@@ -107,16 +219,277 @@ def rc_api_url(tag, arch):
...
+    release_json_str, _ = await exectools.cmd_assert_async(f"oc adm release info {nightly['pullSpec']} -o=json", retries=3)
+    info = nightly["releaseInfo"] = json.loads(release_json_str)
+    nightly["equivalence"] = EquivalenceBase(info)
+
+
+# only look up the same container info once and store it here
+image_info_cache: Dict[str, Dict] = {}
+
+
+class EquivalenceBase:
```
thegreyd, line 237: This looks really close to a Nightly class :D and we could use these objects directly instead of using Dicts? But not a strong opinion..

sosiouxme:
@thegreyd I refactored toward a Nightly class, I think it's cleaner. Also ended up removing the initial RHCOS comparison by labels which wasn't adding much value. However further optimization to limit generation of candidate sets seemed like it would add complexity for not much benefit; it would only reduce the number of release images pulled, but those already happen quickly in parallel. Deeper equivalence checking is already limited. So, no change there.

thegreyd:
Let's ship it :shipit:

PR #625, opened 2022-07-25
Option to disable gssapi_login

ashwindasr:
Option to disable gssapi_login for requests that do not require kerberos keytab


sosiouxme (approved):
/lgtm

PR #627, opened 2022-07-31
[ART-4659] release:gen-payload refactor huge cli method

sosiouxme:
took this huge method and broke it into smaller conceptual units that could be tested. added on one small change, may have one or two more because i think the heterogeneous "moist run" should be dryer (would like to have functional tests that do not actually create new manifest lists constantly)

[Test run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-sync/26629/artifact/gen-payload-artifacts/) produced [this multi nightly](https://multi.ocp.releases.ci.openshift.org/releasestream/4.12.0-0.nightly-multi/release/4.12.0-0.nightly-multi-2022-08-24-034031)

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -759,17 +759,32 @@ class GenPayloadCli:
 
     def sync_heterogeneous_payloads(self, multi_specs: Dict[bool, Dict[str, Dict[str, PayloadEntry]]]):
         """
-        We now generate the artifacts to create heterogeneous release payloads. A heterogeneous or
-        'multi' release payload is a manifest list (i.e. it consists of N release payload
-        manifests, one for each arch). The release payload images referenced in the multi-release
-        payload manifest list are themselves somewhat standard release payloads (i.e. they are
-        based on CVO images for their arch) BUT, each component image they reference is a manifest
-        list.
+        We now generate the artifacts to create heterogeneous release payloads (suitable for
+        clusters with multiple arches present). A heterogeneous or 'multi' release payload is a
+        manifest list (i.e. it consists of N release payload manifests, one for each arch).
 
-        For example, the `cli` image in the s390x release payload will point to a a manifest list
-        composed of cli image manifests for each architecture. In short, the highest level release
-        payload pullspec is a manifest list, referencing single arch release payload images, and
-        these arch specific payload images reference manifest list based components pullspecs.
+        The release payload images referenced in the multi-release payload manifest list are
+        themselves somewhat standard release payloads (i.e. they are based on CVO images for their
+        arch) BUT, each component image they reference is a manifest list. For example, the `cli`
+        image in the s390x release payload will point to a a manifest list composed of cli image
+        manifests for each architecture.
+
+        So the top-level release payload pullspec is a manifest list, referencing release payload
+        images for each arch, and these arch specific payload images reference manifest list based
+        components pullspecs.
+
+        The outcome is that this method creates/updates:
+        1. A component imagestream containing a tag per image component, similar to what we do for
+           single-arch assemblies, but written to a file rather than applied to the cluster.
+        2. For each arch, a payload image based on its CVO, built with references to the (multi-arch)
+           components in this imagestream and stored in
+           * quay.io/openshift-release-dev/ocp-release-nightly for nightlies or
+           * quay.io/openshift-release-dev/ocp-release for standard/custom releases.
+        3. A multi payload image, which is a manifest list built from the arch-specific release
+           payload images and stored in the same quay.io repos as above.
+        4. A _release_ imagestream per assembly to record tag(s) for multi-arch payload image(s)
+           created in step 3, e.g. `4.11-art-latest-multi` for `stream` assembly or
+           `4.11-art-assembly-4.11.4-multi` for other assemblies.
         """
         for private_mode in self.privacy_modes:
 
@@ -785,9 +800,9 @@ class GenPayloadCli:
             imagestream_namespace, imagestream_name = payload_imagestream_namespace_and_name(
                 *self.base_imagestream, "multi", private_mode)
 
-            multi_release_name: str  # The name of the multi release payload image
-            multi_release_manifest_list_tag: str  # The tag for the multi imagestream
-            multi_release_name, multi_release_manifest_list_tag = self.get_multi_release_names(private_mode)
+            multi_release_istag: str  # The tag to record the multi release payload image
+            multi_release_manifest_list_tag: str  # The quay.io tag to preserve the multi payload
+            multi_release_istag, multi_release_manifest_list_tag = self.get_multi_release_names(private_mode)
 
             multi_istags: List[Dict] = list(
                 self.build_multi_istag(tag_name, arch_to_payload_entry, imagestream_namespace)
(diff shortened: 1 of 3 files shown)


doozerlib/cli/release_gen_payload.py
```diff
@@ -58,762 +145,869 @@ def modify_and_replace_api_object(api_obj: oc.APIObject, modifier_func: Callable
...
+        manifest_list_dests: Set[str] = set(entry.dest_manifest_list_pullspec for entry in entries)
+        if len(manifest_list_dests) == 1 and set(manifest_list_dests).pop():  # only one, and not None
+            # Flow 1: Just reuse the manifest list built in brew and synced to a tag in quay.
+            output_digest_pullspec = exchange_pullspec_tag_for_shasum(
+                manifest_list_dests.pop(),
+                entries[0].build_inspector.get_manifest_list_digest())
+            self.logger.info(f"Reusing brew manifest-list {output_digest_pullspec} for component {tag_name}")
+        else:
+            # Flow 2: Build a new manifest list and push it to quay.
+            output_digest_pullspec = self.create_multi_manifest_list(tag_name, arch_to_payload_entry, imagestream_namespace)
```
sosiouxme, line 851: I think we'll also get this flow for those that don't build for all arches, since `pod` will be subbed in for non-building arches. But what if we just used the original manifest list for those too? The reference would be satisfied but never used in the non-building arches, shouldn't that be fine?

doozerlib/cli/release_gen_payload.py
```diff
@@ -58,762 +145,869 @@ def modify_and_replace_api_object(api_obj: oc.APIObject, modifier_func: Callable
...
-            metadata['release.openshift.io/architecture'] = 'multi'
-            new_release_cmd.append("--metadata")
-            new_release_cmd.append(json.dumps(metadata))
-
-            # Create the arch specific release payload containing tags pointing to manifest list component images.
-            exectools.cmd_assert(new_release_cmd)
-
-        # Create manifest list spec containing references to all the arch specific release payloads we've created
-        manifests = []
+        for arch, payload_entry in multi_specs[private_mode]["cluster-version-operator"].items():
```
sosiouxme, line 921: I noticed in a test run limited to two images that if I left out `cluster-version-operator` this failed (of course). But it occurred to me to wonder whether we get a multi payload with only the listed group items. I think so, since there's no way to retrieve the set of tags that were synced in previous runs; well, payload creation probably fails due to missing references but... maybe not if components are included by label instead of reference? There should probably be safeguards against syncing a partial hetero payload.

doozerlib/cli/release_gen_payload.py
jupierce: I feel there is a chance we'd end up breaking something in the ecosystem. Historical releases have guaranteed that if container tooling queries or pulls a release tag, there will be _something_ there for its arch. In the proposed approach, an s390x client may have a manifest list reference to try to inspect/pull, but an error would result if it tried. 

The only reward I can think of for taking the risk is avoiding the creation of several manifest lists in ART's pipeline. If you believe that is a significant enough time savings, perhaps we could turn on the strategy for 4.12 only and see if anything breaks.

sosiouxme: (after discussion)
The point is well taken, nothing should _need_ to look at the missing arch images, but something might be anyway and could break when its expectations aren't met.

My concern was more that we're going to build new manifest lists for these every time we sync, which I'd like to avoid just for hygiene. You came up with a good solution - checksum the manifest list content and tag the manifest list in quay per that checksum; then when a manifest list is called for, look for it under the checksum before creating a new one. (And just for the sake of security, check that the manifest list content does indeed match the desired content, and explode if it doesn't because quay is compromised in that case.)

I think this will be a follow-on improvement.

doozerlib/cli/release_gen_payload.py
sosiouxme: (after discussion)
If we had to, we could look up the existing single-arch imagestreams, stitch them together, and then overlay them with the targeted content updates. However this seems like a lot of work we probably won't need. The simpler solution is to disallow this in the first place. When the image set is limited (either by inclusion or exclusion) and a multi release is requested, raise an error during the doozer parameter validation. Params should either enable a full multi release or not request a multi release.

This may be a follow-on improvement.

doozerlib/cli/release_gen_payload.py
```diff
@@ -59,762 +146,869 @@ def modify_and_replace_api_object(api_obj: oc.APIObject, modifier_func: Callable
...
+
+        with rt.shared_build_status_detector() as bsd:
+            # Use the list of builds associated with the group/assembly to warm up BSD caches
+            bsd.populate_archive_lists(assembly_build_ids)
+            bsd.find_shipped_builds(assembly_build_ids)
+
+        # check that RPMs belonging to this assembly/group are consistent with the assembly definition.
+        for rpm_meta in rt.rpm_metas():
+            self.assembly_issues.extend(assembly_inspector.check_group_rpm_package_consistency(rpm_meta))
+        if rt.assembly == "stream":
```
vfreex, line 331: Per https://issues.redhat.com/browse/ART-3162, it is better not to hardcode the assembly name:

```suggestion
        self.runtime.assembly_type is AssemblyTypes.STREAM
```

doozerlib/cli/release_gen_payload.py
```diff
@@ -59,762 +146,869 @@ def modify_and_replace_api_object(api_obj: oc.APIObject, modifier_func: Callable
...
-        assembly_issues.append(mismatch_issue)
+    def detect_non_latest_rpms(self, assembly_inspector: AssemblyInspector):
+        """
+        If this is a stream assembly, images which are not using the latest rpm builds should not reach
+        the release controller. Other assemblies may be deliberately constructed from non-latest.
+        """
+        self.logger.debug("detecting images with group RPMs installed that are not the latest builds...")
+        for dgk, build_inspector in assembly_inspector.get_group_release_images().items():
+            if build_inspector:
+                # dgk = build_inspector.get_image_meta().distgit_key  #XXX: check if there was some reason to do this
```
vfreex, line 438: I believe this is a leftover which can be removed.

doozerlib/cli/release_gen_payload.py
```diff
@@ -59,762 +146,869 @@ def modify_and_replace_api_object(api_obj: oc.APIObject, modifier_func: Callable
...
+        """
+        Associate relevant assembly issues with an RHCOS PayloadEntry.
+        Use assembly inspector to detect assembly issues like RPMs installed in
+        the PayloadEntry that are not what the assembly specifies.
+        """
+        # record issues found in assembly list and per payload entry
+        self.assembly_issues.extend(assembly_inspector.check_rhcos_issues(payload_entry.rhcos_build))
+        payload_entry.issues.extend(ai for ai in self.assembly_issues if ai.component == "rhcos")
+
+        if self.runtime.assembly == "stream":
```
vfreex, line 517: Same here.

doozerlib/cli/release_gen_payload.py
```diff
@@ -59,762 +146,869 @@ def modify_and_replace_api_object(api_obj: oc.APIObject, modifier_func: Callable
...
+            # For normal 4.x-art-latest, we update the imagestream with individual component images
+            # and the release controller formulates the nightly. For multi-arch, this is not
+            # possible (notably, the CI internal registry does not support manifest lists). Instead,
+            # in the ocp-multi namespace, the 4.x-art-latest imagestreams are configured
+            # `as: Stable`: https://github.com/openshift/release/pull/24130
+            # This means the release controller treats entries in these imagestreams the same way it
+            # treats it when ART tags into is/release; i.e. it treats it as an official release.
+            # With this comes the responsibility to prune nightlies ourselves.
+            release_tags: List = obj_model.spec["tags"]
+            while len(release_tags) > 5:  # keep at most 5 tags
```
vfreex, line 986: Per https://coreos.slack.com/archives/GDBRP5YJH/p1661329456656149, we only keep 1 tag for non-stream assemblies.

vfreex (changes requested):
Great work @sosiouxme ! I went through the whole logic and didn't noticed anything is missing comparing to the current commit. There is one thing about image stream tags for non-stream multi payload.

sosiouxme:
my force-push was just a history rewrite with no changes, you'll have to trust me on that :) going to merge now.

PR #628, opened 2022-08-03
Add timeouts to oc image mirror

vfreex:
We observed an issue that `oc image mirror` had stuck for 2 hours.

Add a timeout of 30 minutes to each `oc image mirror` invocation.

See https://coreos.slack.com/archives/GDBRP5YJH/p1659501833732749.


sosiouxme (approved):
i have doubts about whether retrying will really help much (since we don't know what went wrong - quay.io performance? buildvm io?), but if it takes longer than that then _something_ is wrong so i guess it's worth a try.

PR #630, opened 2022-08-03
release_gen_assembly.py option to output to file

ashwindasr:
Option to write the output of gen-assembly doozer command to a file. Useful for https://github.com/openshift/aos-cd-jobs/pull/3140 as its better to read from a file rather than the output to create the PR.

### DIFF
--- a/doozerlib/cli/release_gen_assembly.py
+++ b/doozerlib/cli/release_gen_assembly.py
@@ -46,7 +46,7 @@ def releases_gen_assembly(ctx, name):
 @click.option("--suggestions-url", metavar='SUGGESTIONS_URL', required=False,
               default="https://raw.githubusercontent.com/openshift/cincinnati-graph-data/master/build-suggestions/",
               help="When using --auto-previous, set custom suggestions URL, load from {major}-{minor}-{arch}.yaml")
-@click.option('--output-file', required=False, help='Specify a file path to write the doozer output to')
+@click.option('--output-file', '-o', required=False, help='Specify a file path to write the doozer output to')
 @pass_runtime
 @click.pass_context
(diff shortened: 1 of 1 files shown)


doozerlib/cli/release_gen_assembly.py
```diff
@@ -46,9 +46,10 @@ def releases_gen_assembly(ctx, name):
 @click.option("--suggestions-url", metavar='SUGGESTIONS_URL', required=False,
               default="https://raw.githubusercontent.com/openshift/cincinnati-graph-data/master/build-suggestions/",
               help="When using --auto-previous, set custom suggestions URL, load from {major}-{minor}-{arch}.yaml")
+@click.option('--output-file', required=False, help='Specify a file path to write the doozer output to')
```
thegreyd, line 49: "to write the generated assembly definition to" instead of doozer output

thegreyd: Also we can add a shortflag `-o`

ashwindasr: Done!

doozerlib/cli/release_gen_assembly.py
```diff
@@ -398,3 +399,7 @@ def exit_with_error(msg):
     }
 
     print(yaml.dump(assembly_def))
+
+    if output_file:
+        with open(output_file, 'w') as file:
```
thegreyd, line 404:
```suggestion
        stream = file(output_file, 'w')
        yaml.dump(data, stream)
```

thegreyd: Essentially the same but more concise

ashwindasr: Done!

PR #632, opened 2022-08-04
get-nightlies: Allow skipping deeper nightly comparison checks

thegreyd:
This is useful in getting nightlies for dev-previews 
when we don't need 100% consistency
[slack context](https://coreos.slack.com/archives/CB95J6R4N/p1659645812957009?thread_ts=1659634736.706589&cid=CB95J6R4N)
```
doozer -q -g openshift-4.12 get-nightlies --allow-inconsistency

4.12.0-0.nightly-s390x-2022-08-01-145737 4.12.0-0.nightly-ppc64le-2022-08-01-145640 4.12.0-0.nightly-arm64-2022-08-01-174630 4.12.0-0.nightly-2022-08-01-151317
```

### DIFF
--- a/doozerlib/cli/get_nightlies.py
+++ b/doozerlib/cli/get_nightlies.py
@@ -15,6 +15,7 @@ logger = logutil.getLogger(__name__)
 
 @cli.command("get-nightlies", short_help="Determine set(s) of accepted nightlies with matching contents for all architectures")
 @click.option("--matching", metavar="NIGHTLY_NAME", multiple=True, help="Only report nightlies with the same content as named nightly")
+@click.option("--allow-inconsistency", is_flag=True, help="Allow nightlies that fail deeper consistency checks")
 @click.option("--allow-pending", is_flag=True, help="Include nightlies that have not completed tests")
 @click.option("--allow-rejected", is_flag=True, help="Include nightlies that have failed tests")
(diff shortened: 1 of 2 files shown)


doozerlib/cli/get_nightlies.py
```diff
@@ -104,17 +109,24 @@ async def get_nightlies(runtime, matching: List[str], exclude_arch: List[str], a
...
             util.green_print(nightly_set.details() if details else nightly_set)
-            if len(nightly_sets) >= limit:
+            if len(consistent_nightly_sets) >= limit:
                 break  # don't spend time checking more than were requested
 
-    if not nightly_sets:
+    remaining = limit - len(consistent_nightly_sets)
+    if allow_inconsistency and remaining > 0:
+        for nightly_set in nightly_sets[:remaining]:
+            util.yellow_print(nightly_set.details() if details else nightly_set)
```
sosiouxme, line 126: one minor nit with this is that `nightly_sets` includes `consistent_nightly_sets` so consistent ones will be double-printed if the limit is high enough. We're unlikely to run this with a limit higher than 1 or 2 so it's not exactly important but I just like to be thorough. Adding a commit that I think simplifies this.

sosiouxme (approved):

PR #633, opened 2022-08-05
ART-4334 Change default multi payload repo to ocp-release-nightly

thegreyd:
https://issues.redhat.com/browse/ART-4334


sosiouxme (approved):
hopefully the pullspec is fed to the release-controller and not hardcoded somewhere instead...

PR #643, opened 2022-09-16
[ART-4751] update valid-subscription in csv

Ximinhan:
Reverts openshift/doozer#642

### DIFF
--- a/doozerlib/olm/bundle.py
+++ b/doozerlib/olm/bundle.py
@@ -254,7 +254,14 @@ class OLMBundle(object):
                 contents = self.find_and_replace_image_references_by_sha(f.read())
                 f.seek(0)
                 f.truncate()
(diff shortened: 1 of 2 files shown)


thiagoalessio (commented):
/hold

It was confirmed by @lance that the label should be present in the CSV and not in `annotations.yaml`. Which is accomplished by this code change. A side-effect of this PR, is that the label will be gone from the Dockerfile as well, is that correct/desired?

Example: https://pkgs.devel.redhat.com/cgit/containers/cluster-nfd-operator-bundle/tree/Dockerfile?h=rhaos-4.11-rhel-8#n23 Last line will be removed if this PR gets merged.

Ximinhan:
confirmed from email, no need for Dockerfile, so the pr is still valid

PR #644, opened 2022-09-16
gen-payload: remove old tags from multi imagestream for non-stream as…

vfreex:
…sembly

We only need to keep 5 old imagestreamtags for stream assembly. Non-stream assemblies should contain one single tag.

For more info, see https://coreos.slack.com/archives/GDBRP5YJH/p1661329456656149

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -1001,18 +1001,23 @@ class GenPayloadCli:
             if obj_model.spec.tags is oc.Missing:
                 obj_model.spec["tags"] = []
 
-            # For normal 4.x-art-latest, we update the imagestream with individual component images
-            # and the release controller formulates the nightly. For multi-arch, this is not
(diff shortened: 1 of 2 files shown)


vfreex:
> I was working on this same change per our discussion; but in my case I was also recording the tags that got removed, thinking we could prune them from the imagestream status (not just the spec) and maybe from the quay release repo too. But that could just be a follow-on to this.

Status is usually managed by the controller, so I don't think we can directly prune it. But I guess it may have a GC mechanism? 

For quay repo tags, we are missing the permission to prune them. Maybe we can use deterministic quay tags so that a later gen-payload run will always override the former one?

The direct reason for having this PR is that promote-assembly job [expects only one single tag in the imagestream](https://coreos.slack.com/archives/C02TN2SJR0F/p1663250243434359?thread_ts=1663250030.098869&cid=C02TN2SJR0F), otherwise it doesn't know which tag to promote (or maybe we can introduce another mechanism to determine which tag is the latest one, but that adds unnecessary complex and is not future-proof).

PR #652, opened 2022-09-24
Speed up tests

joepvd:
From: `Ran 234 tests in 8.211s`
To: `Ran 234 tests in 1.299s`


sosiouxme:
/lgtm
talk about trimming out the deadwood...

PR #655, opened 2022-09-28
use actual build nvr

Ximinhan:


doozerlib/osbs2_builder.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         self.task_id: int = 0
         self.task_url: str = ""
         self.nvr: str = ""
```
Ximinhan, line 32: this nvr is using the name from distant not brew, this will make the build name in the record passed into olm bundle job incorrect

sosiouxme: i like getting the nvr from the build record. what i don't like is that the wrong nvr is still set up the `build()` method below and someone might use it without noticing it's wrong.

Ximinhan:
other member will fix/refactor this, so close this one

PR #658, opened 2022-09-30
[ART-4861] rpms:rebase: Allow to use `~` in rpm version

vfreex:
So something like `4.12.0~rc.1` will become a valid rpm version.


joepvd (approved):
/lgtm

PR #660, opened 2022-10-04
Log database errors and fix alter table issue

ashwindasr:
If a new label is added in the Dockerfile of an Openshift component, doozer should add a column in the mariadb database table. But it seems like it hasn't been doing so, for any version. Because of the error, no logs are being written for that component. For example the last successful hypershift build log written to the db was in May, before they added [new labels in the docker file](https://github.com/openshift/hypershift/commit/5d49ed41f5cbeb3c656e0727029c642d7e0dadca). Till now the error wasn't being caught/logged.

Update:
After using this doozer fork to run a [build](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fcustom/2764/consoleFull) to catch the error, it was found that the problem was because the new column name was too long. SQL column names cannot be greater than 64 characters.
```
2022-10-06 00:04:03,145 ERROR Database error: 1059 (42000): Identifier name 'label_io_openshift_hypershift_control_plane_operator_skips_haproxy' is too long
```

Fix:
If the column name is greater than 64, hash, truncate and append the hash.


ashwindasr:
It was decided that its better to use an allow list to store only the necessary labels and skip the others, otherwise the table would grow unnecessarily large.

PR #663, opened 2022-10-05
Handle exceptions during upstream equivalent rebase

locriandev:
Context: https://coreos.slack.com/archives/GDBRP5YJH/p1664989685327059?thread_ts=1664955491.116099&cid=GDBRP5YJH


sosiouxme (approved):
/lgtm

PR #668, opened 2022-10-10
Allow list for io.openshift labels

ashwindasr:
If a new label is added in the Dockerfile of an Openshift component, doozer should add a column in the mariadb database table. But it seems like it hasn't been doing so, for any version. Because of the error, no logs are being written for that component. For example the last successful hypershift build log written to the db was in May, before they added [new labels in the docker file](https://github.com/openshift/hypershift/commit/5d49ed41f5cbeb3c656e0727029c642d7e0dadca). Till now the error wasn't being caught/logged.

After using this doozer fork to run a [build](https://saml.buildvm.openshift.eng.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fcustom/2764/consoleFull) to catch the error, it was found that the problem was because the new column name was too long. SQL column names cannot be greater than 64 characters.
```
2022-10-06 00:04:03,145 ERROR Database error: 1059 (42000): Identifier name 'label_io_openshift_hypershift_control_plane_operator_skips_haproxy' is too long
```

Fix:
- Use an allow list for `io.openshift` labels
- If label length exceed 64 characters, log it and skip


doozerlib/dblib.py
```diff
@@ -327,12 +327,17 @@ def handle_missing_columns(self, payload, table_name):
...
+                    cursor.execute(f"alter table {table_name} add column `{missing_column}` {column_type}")
+                    self._table_column_cache[table_name][missing_column] = True
+
+                    self.runtime.logger.info("Added new column [{}] of identified type [{}] to table [{}] "
+                                             "of database [{}].".format(missing_column, column_type,
+                                                                        table_name, self.db))
+                except Exception as e:
+                    self.runtime.logger.warning("Could not add new column [{}] of identified type [{}] to table [{}] "
+                                                "of database [{}]. Error: {}".format(missing_column, column_type,
+                                                                                     table_name, self.db, e))
```
sosiouxme, line 340: at this point, having identified a column that we can't write, can we remove that column from the payload so that the rest of it can still be written? i'm expecting that the next thing you get after this warning is a SQL error trying to write to an undefined column.

doozerlib/dblib.py
```diff
@@ -470,6 +475,14 @@ def __init__(self, db, table, extras={}, dry_run=False):
 
         self.attrs.update(extras)
 
+        # An allow list for labels that start with io.openshift
+        self.io_label_allow_list = [
+            'label_io_openshift_build_source_location',
+            'label_io_openshift_build_commit_id',
+            'label_io_openshift_build_commit_url',
+            'label_io_openshift_release_operator',
+            'label_io_openshift_build_versions']
```
sosiouxme, line 484: yeah... don't really want this here to affect every future use of `Record`....

functional_tests/test_database_write.py
```diff
@@ -0,0 +1,162 @@
...
+
+        self.db.connection = mysql.connector.connect(host='localhost',
+                                                     database='art_dash',
+                                                     user='root',
+                                                     password='secret')
+        self.db.mysql_db_env_var_setup = True  # since we're setting the db connection above
+        self.db.db = "art_dash"
+        self.table_name = "log_build"
+
+    def test_1(self):
```
sosiouxme, line 108: in general, good to have more descriptive test method names so that when it fails you recognize it...

sosiouxme (changes requested):

doozerlib/dblib.py
ashwindasr: Actually there's no error raised, just a Null value is entered in the DB, same as if we remove it from the payload. That's why it slipped my mind. But its better to delete it from the payload like you said.

doozerlib/dblib.py
ashwindasr: Changed to distgit.py

sosiouxme (approved):

sosiouxme:
:shipit:

PR #676, opened 2022-11-02
fix https_proxy value is string null

Ximinhan:
jenkins job change empty value to string null when passing the value as arg to doozer


vfreex (changes requested):
'null' is just a normal string. It should be fixed in jenkins rather than doozer.

Ximinhan:
ok

PR #677, opened 2022-11-02
Add back "viable" in gen-payload report

joepvd:


doozerlib/cli/release_gen_payload.py
```diff
@@ -270,7 +270,7 @@ def run(self):
         self.payload_entries_for_arch = self.generate_payload_entries(assembly_inspector)
         assembly_report: Dict = self.generate_assembly_report(assembly_inspector)
 
-        print(yaml.dump(assembly_report["assembly_issues"], default_flow_style=False, indent=2))
+        print(yaml.dump(assembly_report, default_flow_style=False, indent=2))
```
thegreyd, line 273: https://github.com/openshift/doozer/blob/c4fb0baa1a96424ad999c9dcfce75a57f2b65469/doozerlib/cli/release_gen_payload.py#L305 assembly_report contains more things than issues and viable.. are we sure we want to print those images?

vfreex (approved):
/lgtm

PR #686, opened 2022-11-15
Assembly: allow to pin rpm dependencies in any assembly

vfreex:
Currently only assemblies with a basis event allow pins. Remove this restriction to allow to pin specific rpms for stream.


thiagoalessio (approved):
/lgtm

PR #688, opened 2022-11-17
get-nightlies: honor --arches global option

vfreex:
This is useful to get a consistent set of nightlies for microshift rebase because we only care about certain arches.

```sh
doozer --arches=x86_64,aarch64 --group=openshift-4.12 --assembly=stream get-nightlies --matching=4.12.0-0.nightly-2022-11-16-003434 --limit=2
```

Use `runtime.arches` to retrieve all enabled arches instead of reading group config.

### DIFF
--- a/doozerlib/cli/get_nightlies.py
+++ b/doozerlib/cli/get_nightlies.py
@@ -1,14 +1,15 @@
-from typing import List, Dict, Optional, Set, Tuple
 import asyncio
-import click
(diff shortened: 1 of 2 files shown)


vfreex:
poll-payload job already passes `--arches` option to get-nightlies. So once this is merged and deployed, we will get new nightly microshift builds automatically.

PR #689, opened 2022-11-17
Use art-latest imagestream only for assembly 'stream'

locriandev:
Context: https://coreos.slack.com/archives/GDBRP5YJH/p1668614858651439


vfreex (approved):
/lgtm

PR #696, opened 2022-11-25
WIP Art 5258 Consume rhcos multi builds

thegreyd:
Right now in gen-payload we go through each arch to 
fetch latest rhcos build ids and pullspecs since they were
separate builds. Now with multi builds, there is a single build_id.
Before to fetch a latest build we could go to builds.json and see
an entry there and assume the build was ready to be consumed,
but in the multi-world, builds.json is populated with a build_id as
soon as the build starts, so it can be "in progress", "failed" or "success",
In case of "in progress" or "failed", the data simply won't exist in the 
metadata files. The way to be sure that all arches are complete is to 
query quay for the build_id, check the manifest list for all arches

```
oc image info -o json quay.io/openshift-release-dev/ocp-v4.0-art-dev:412.86.202211241304-0

OS            DIGEST
  linux/amd64   sha256:050066bd655a5e9573bd88be52d11b7c60289b0e32407bc13222876fc8739ff4
  linux/s390x   sha256:f2d7eb6e87f38bd9bd8e635188a272fb5c50992faadc19913ffdb104d7fd31db
  linux/ppc64le sha256:cce70b72f17d65160f4d62410ab0c2e178f312da776f45428e9aa6b63a8491af
```

So we want to move away from fetching rhcos by individual arches
to fetching rhcos by being aware of all arches, and then selecting the
build_id that matches the criteria.

### DIFF
--- a/doozerlib/assembly_inspector.py
+++ b/doozerlib/assembly_inspector.py
@@ -1,4 +1,4 @@
-from typing import List, Dict, Optional
+from typing import List, Dict, Optional, Iterable
 
 from koji import ClientSession
 from doozerlib.rpm_utils import parse_nvr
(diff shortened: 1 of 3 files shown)


sosiouxme:
Don't think we're going to go this route.

PR #700, opened 2022-12-01
Refactor gen assembly

locriandev:


tests/cli/test_gen_assembly.py
```diff
@@ -0,0 +1,177 @@
...
+        gacli = GenAssemblyCli(runtime=runtime, gen_assembly_name='rc.0')
+        self.assertEqual(gacli.assembly_type, AssemblyTypes.CANDIDATE)
+
+        gacli = GenAssemblyCli(runtime=runtime, gen_assembly_name='fc.0')
+        self.assertEqual(gacli.assembly_type, AssemblyTypes.CANDIDATE)
+
+        gacli = GenAssemblyCli(runtime=runtime, gen_assembly_name='ec.0')
+        self.assertEqual(gacli.assembly_type, AssemblyTypes.PREVIEW)
+
+        gacli = GenAssemblyCli(runtime=runtime, gen_assembly_name='just-a-name')
```
joepvd, line 28: I believe I recall that an assembly with `^[0-9]+\.[0-9]+\.[0-9]+` should be standard, and `custom` assemblies should match `^art[0-9]+$`, and that strings other than `stream`, `test`, and `^[efr]c\.[0-9]+$` should raise an error. Or maybe that is enforced some place else?

tests/cli/test_gen_assembly.py
```diff
@@ -0,0 +1,177 @@
...
+from unittest.mock import MagicMock, Mock, patch
+
+from flexmock import flexmock
+
+import doozerlib
+from doozerlib.assembly import AssemblyIssueCode, AssemblyTypes, AssemblyIssue
+from doozerlib.cli.release_gen_assembly import GenAssemblyCli
+
+
+class TestGenPayloadCli(TestCase):
```
joepvd, line 11: Really love the test coverage, and how clean it looks!

Could you maybe add a docstring to the `test_` methods explaining what business logic this is covering? I noticed that method name of the test and the input and the assertion did not always make it clear for me what this is really about.

joepvd (commented):
Very nice! I love how clean the tests look, and how comprehensive the coverage is. 

Left an optional suggestion of improvement to communicate somehow what the business logic is in the comments, but feel free to ignore. 
And I am not sure about what should happen with assembly named `just-a-name`, maybe you could clarify?

tests/cli/test_gen_assembly.py
locriandev: There's confusion between assembly name and gen-assembly name. `--assembly` must always set to `stream`, [or the command will fail](https://github.com/openshift/doozer/blob/8b52cabb1a3bb12329ca134260df95b3879c8c45/doozerlib/cli/release_gen_assembly.py#L169).

The `--name` option indicates the name of the assembly (e.g. "4.9.99", "art1234") to scaffold. The logic according to which the assembly type is deduced is [here](https://github.com/openshift/doozer/blob/8b52cabb1a3bb12329ca134260df95b3879c8c45/doozerlib/cli/release_gen_assembly.py#L141).

In order for a gen-assembly name to be of type `custom`, the `--custom` option must be specified from the CLI

joepvd (approved):
/lgtm

PR #701, opened 2022-12-03
bundle: Ensure referenced images are declared as dependencies

joepvd:
...and if not, fail during rebase


joepvd:
Tested with https://github.com/openshift/ocp-build-data/pull/2325 and https://github.com/openshift/ocp-build-data/pull/2326

PR #702, opened 2022-12-03
Operator rebase: Actually find image-references

joepvd:


doozerlib/distgit.py
```diff
@@ -2083,6 +2083,7 @@ def _get_csv_file_and_refs(self, csv_config):
         for cand in ref_candidates:
             if os.path.isfile(cand):
                 refs = cand
+                break
```
locriandev, line 2086: It seems to me that adding this `break` statement just interrupts the loop as soon as the first candidate is found. Without it, it would overwrite `refs` as new candidates come along. 
Could you provide some context about what we're solving here?

PR #703, opened 2022-12-04
Instrument images:rebase with database

joepvd:


locriandev:
Is `images:update-dockerfile` never used anywhere? If so, I'd make it explicit we're removing it and why

locriandev:
> @locriandev `images:update-dockerfile` does not get the latest news from GitHub, so it keeps upstream sources the same in distgit. It just updates the dockerfile to enable triggering a new build.
> 
> This option is not used in automation. It used to have an important function when signing of rpms was not yet automated: It would help to rebuild the exactly same images but now with signed rpms.
> 
> You can still use it in the custom job (but I think nobody uses that functionality).
> 
> Think it'd be fine to remove that functionality, but maybe not in the scope of this PR.

I've created https://github.com/openshift/doozer/pull/719 to make it explicit we want to remove `images:update-dockerfile`

doozerlib/cli/images_rebase.py
```diff
@@ -69,48 +69,61 @@ def images_rebase(runtime: Runtime, version: Optional[str], release: Optional[st
     metas = runtime.ordered_image_metas()
     lstate['total'] = len(metas)
 
-    def dgr_rebase(image_meta, terminate_event):
-        try:
+    def dgr_rebase(image_meta, terminate_event, runtime):
+        with runtime.db.record('rebase', metadata=image_meta) as db_entry:
+            db_entry.update({'state': 'unstarted', 'embargoed': str(embargoed)})
             dgr = image_meta.distgit_repo()
+            db_entry.update({'upstream_repo': dgr.config.content.source.git.url or '', 'upstream_branch': dgr.config.content.source.git.branch.target or ''})
```
vfreex, line 76: `image_meta.distgit_repo()` can raise an exception if it fails to clone the distgit repo. This needs to be wrapped in a `try` block otherwise the exception clause will be captured (currently we have the logic to update record.log on excpetion). in such a case a simple record like `{'state': 'unstarted', 'embargoed': str(embargoed)}` won't help much in my opinion.

doozerlib/cli/images_rebase.py
```diff
@@ -69,48 +69,61 @@ def images_rebase(runtime: Runtime, version: Optional[str], release: Optional[st
     metas = runtime.ordered_image_metas()
     lstate['total'] = len(metas)
 
-    def dgr_rebase(image_meta, terminate_event):
-        try:
+    def dgr_rebase(image_meta, terminate_event, runtime):
```
vfreex, line 72: `runtime` is in the scope. Why adding another parameter here?

vfreex (changes requested):
A few minor issues. Otherwise lgtm

locriandev:
https://issues.redhat.com/browse/ART-7329 has been opened to track this work.

PR #704, opened 2022-12-05
Fix "task failed, build succeeded" situation in osbs2

joepvd:
It can happen that a brew task fails, but brew considers the build to be good. This commit moves the calculation of the resulting nvr forward, and upon task failure, asks brew if the nvr has been built successfully.

### DIFF
--- a/doozerlib/osbs2_builder.py
+++ b/doozerlib/osbs2_builder.py
@@ -60,6 +60,7 @@ class OSBS2Builder:
 
         error = None
         message = None
+        nvr = f"{image.get_component_name()}-{dg.org_version}-{dg.org_release}"
         for attempt in range(retries):
             logger.info("Build attempt %s/%s", attempt + 1, retries)
             try:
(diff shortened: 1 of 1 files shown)


thiagoalessio (approved):
/lgtm

sosiouxme:
just a couple notes before merging:

1. There are cases where don't (and don't want to) know the NVR before the build is complete: when the `release` is not defined, brew will pick one that auto-increments. We used to do this in OCP 3 but haven't in 4, and no longer do in 3. The only place where we do this now is olm bundles, I'd want to check that this does not break that.
2. doozer can do `scratch` builds which won't get a build record. We don't use this much (partly because doozer's grasp of this situation is lacking... with no NVR it already considers these builds failed). Probably not important here, just noting it's a violation of the assumption that there will be an NVR resulting from the build.

It seems we're pretty committed to specifying `release` (outside olm bundles) and don't use scratch builds. Probably code enabling these use cases should either be removed or made fully-functional.

I'm fine with merging though.

vfreex (approved):
/lgtm

PR #710, opened 2022-12-13
Expose jira information from components

jupierce:
- Migrate old bugzilla logic to be Jira based. 
- Consults prodsec database to determine component to Jira component mapping.
- Adds a new `hosts` section in doozer/settings.yaml to let us alias hostnames without checking them into git (avoid getting flagged by prodsec)

### DIFF
--- a/Usage.md
+++ b/Usage.md
@@ -25,6 +25,10 @@ group: openshift-4.0
 #Username for running rhpkg / brew / tito
 user: ahaile
 
+# URLs for internal systems
+hosts:
+  prodsec_git: ...
+
 #Global option overrides that can only be set from settings.yaml
 global_opts:
   # num of concurrent distgit pull/pushes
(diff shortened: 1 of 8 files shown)


jupierce:
@vfreex - `Unknown` is an actual component name in the [OCPBUGS](https://issues.redhat.com/projects/OCPBUGS) project in Jira. So the worst case scenario for a component that is slow to add to the prodsec database should be bugs filed against `Unknown` . This is not uncommon and they should be routed by someone familiar with the product.

jupierce:
> If data in prodsec's repo doesn't change very frequently, maybe we can periodically sync the data into ocp-build-data?
I don't know how often it changes, but a periodic sync would probably work. We'd want to make certain we only pull in a known part of the file and that the semantics don't change. I don't want to accidentally pull in secret prodsec data.

vfreex:
> I don't know how often it changes, but a periodic sync would probably work. We'd want to make certain we only pull in a known part of the file and that the semantics don't change. I don't want to accidentally pull in secret prodsec data.

Makes sense to me. I am ok with merging this and consider to do periodic sync when there is a need.

/lgtm

PR #714, opened 2022-12-15
[ART-3683] validate rhcos <-> driver-toolkit kernel consistency

sosiouxme:
Add the ability to configure package consistency requirements between RHCOS and other payload members which are validated during `release:gen-payload`.
https://github.com/openshift/ocp-build-data/pull/2372 shows intended config for 4.13

Also added a relatively "quick" check via `inspect:stream FAILED_CONSISTENCY_REQUIREMENT`
And finally, this gets rid of the excess AssemblyIssues in releases where there are multiple RHCOS payload entries - with this change they are only assessed once per arch (only for the "primary" container). This will also help in the near future when we have an RHCOS build for RHEL 8 and RHEL 9 - will not attempt to compare them to each other.


doozerlib/assembly.py
```diff
@@ -14,15 +14,52 @@ class AssemblyTypes(Enum):
...
+    # so there is a risk of not officially shipping the RPM we are using, or
+    # of misreporting bug fixes. We may either rebuild the container or change
+    # the assembly to resolve.
     OUTDATED_RPMS_IN_STREAM_BUILD = 4
+
+    # The arch-specific builds of RHCOS did not all install the same RPM
+    # versions (if they installed at all).
     INCONSISTENT_RHCOS_RPMS = 5
+
+    # A dependency was defined explicitly for an assembly member, but it is not    # installed, i.e. the ARTists expected something to be installed, but it
```
joepvd, line 45: newline

doozerlib/assembly.py
```diff
@@ -14,15 +14,52 @@ class AssemblyTypes(Enum):
 
 
 class AssemblyIssueCode(Enum):
+
```
joepvd, line 17: I really really really like these comments 😍

joepvd (approved):
/lgtm

spotted one typo

PR #715, opened 2022-12-19
scan-sources: Detect member in from.builder

joepvd:
`enterprise-console-builder` has this in their config:

```yaml
from:
  builder:
  - stream: golang
  - member: openshift-base-nodejs
  member: openshift-enterprise-base
```

The code so far did not look for `member` in `.from.builder`. This means that in the case of a rebuild of `openshift-base-nodejs` and `openshift-enterprise-console`, where nodejs succeeded but console failed, scans-sources will not discover that a rebuild is necessary.

With this patch, the following invocation says a rebuild is needed:
```sh
doozer -g openshift-4.12 --assembly stream -i openshift-enterprise-console -i ose-metallb -i openshift-base-nodejs config:scan-sources
```


thiagoalessio (approved):
/lgtm

PR #717, opened 2023-01-06
fix get_maintainer_info is undefined error

vfreex:
This method has been removed by
https://github.com/openshift/doozer/pull/710/.

This PR changes it to the new method `get_jira_info`. Or we can completely remove the logic of notifying maintainers for missing component if this is not needed any more.

### DIFF
--- a/doozerlib/distgit.py
+++ b/doozerlib/distgit.py
@@ -2453,8 +2453,8 @@ class ImageDistGitRepo(DistGitRepo):
             owners = list(self.config.owners)
 
         # If upstream has not identified the BZ component, have the pipeline send them a nag note.
-        maintainer = self.metadata.get_maintainer_info()
-        if owners and not maintainer.get('component', None):
+        jira_project, jira_component = self.metadata.get_jira_info()
+        if jira_project == 'OCPBUGS' and jira_component == 'Unknown':
             week_number = date.today().isocalendar()[1]  # Determine current week number
             oit_path = dg_path.joinpath('.oit')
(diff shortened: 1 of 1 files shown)


doozerlib/distgit.py
```diff
@@ -2453,8 +2453,8 @@ def _merge_source(self):
             owners = list(self.config.owners)
 
         # If upstream has not identified the BZ component, have the pipeline send them a nag note.
-        maintainer = self.metadata.get_maintainer_info()
-        if owners and not maintainer.get('component', None):
+        jira_project, jira_component = self.metadata.get_jira_info()
+        if jira_project == 'OCPBUGS' and jira_component == 'Unknown':
```
Ximinhan, line 2457: do we still need to check owners? 
the component can also be set "none"

Ximinhan, line 2457: do we still need to check owners? 
the component can also be set "none"

vfreex: > do we still need to check owners?

Good point. The job checking for `bz_notify` also checks `owners` before notifying the owners.
I think the upside of checking owners here is a defensive measure, but the downside is we lose the ability to manually analyze with `grep`.

> the component can also be set "none"

Could you give me an exact example?

jupierce:
Sorry for missing this. This check / nag can be completely removed for now.

jupierce:
I've gone ahead a removed the code here since ocp4 jobs are broken: https://github.com/openshift/doozer/pull/718

PR #719, opened 2023-01-16
Remove command 'images:update-dockerfile'

locriandev:
CLI command `images:update-dockerfile` is not used anywhere by automation and can be removed


vfreex:
@sosiouxme Could you confirm this is no longer used?

PR #725, opened 2023-01-31
Migrate away from prodsec data for jira

jupierce:
Prodsec has requested they be dowstream of component Jira mapping data instead of the source of truth. The source of truth has historically been OWNERS files in component repos, but has now been established in https://github.com/openshift/ocp-build-data/blob/main/product.yml .


vfreex (approved):
/lgtm

PR #727, opened 2023-02-02
[ART-5969] Ensure two accepted payloads remain in multi release controller

jupierce:
Unlike single arch release controllers, which prune nightly release payloads themselves, ART is responsible for pruning heterogeneous nightlies. This is due to the historical lack of manifest list support in imagestreams and in oc (this should improve over time).
Until then, ART prunes nightlies and our logic should ensure that at least two accepted release payloads remain in the final list (if there were accepted payloads to begin with). These accepted payloads allow new payloads to have baselines from which to execute upgrade tests.


vfreex (approved):
/lgtm

PR #732, opened 2023-02-23
Use https where possible

joepvd:


.editorconfig
```diff
@@ -1,4 +1,4 @@
-# http://editorconfig.org
+# https://editorconfig.org
```
vfreex, line 1: I think we'd better to keep those irrelevant `http` links.

vfreex (changes requested):

doozerlib/util.py
```diff
@@ -130,7 +130,7 @@ def convert_remote_git_to_https(source_url: str):
     :return: Normalized https git URL
     """
     url = source_url.strip().rstrip('/')
-    url = remove_prefixes(url, 'http://', 'https://', 'git://', 'git@', 'ssh://')
```
thegreyd, line 133: This is actually needed

PR #734, opened 2023-02-27
fix snyk import error

Ximinhan:
fix issue similar to https://github.com/openshift/elliott/pull/513

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,11 @@
 .PHONY: venv tox lint test install
 
 install:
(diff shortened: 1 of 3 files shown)


locriandev:
/hold

see https://redhat-internal.slack.com/archives/C03U0TPBQMR/p1677489370591869?thread_ts=1677482754.791689&cid=C03U0TPBQMR

locriandev (changes requested):
Need to update https://github.com/openshift/doozer/blob/master/tox.ini#L5 as well

PR #737, opened 2023-03-06
rpms:build: Fix hotfix tagging for multiple targets

vfreex:
Some packages, like `microshift`, has 2 targets (one for rhel 8 and another for rhel 9). For non-stream build, we need to tag the resulting builds into `rhaos-4.y-rhel-z-hotfix` to prevent gabbage collection, however Brew no longer allows us to use a single `-hotfix` tag.

With this PR, builds will be tagged into target-specific hotfix tags, e.g. `rhaos-4.13-rhel-9-hotfix` and `rhaos-4.13-rhel-8-hotfix` in order to work around this limitation.

### DIFF
--- a/doozerlib/metadata.py
+++ b/doozerlib/metadata.py
@@ -268,6 +268,12 @@ class Metadata(object):
     def candidate_brew_tags(self):
         return [self.candidate_brew_tag()]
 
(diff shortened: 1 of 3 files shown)


joepvd (approved):
/lgtm

PR #738, opened 2023-03-07
[ART-6053] rhcos: ignore RPMs missing from brew

sosiouxme:
[test run](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/aos-cd-builds/job/build%252Fbuild-sync/31280/console)
enables https://github.com/openshift/ocp-build-data/pull/2619

### DIFF
--- a/doozerlib/rhcos.py
+++ b/doozerlib/rhcos.py
@@ -358,6 +358,7 @@ class RHCOSBuildInspector:
                     rpm_def = koji_api.getRPM(nvra, strict=True)
                 except koji.GenericError as e:
                     if self.runtime.group_config.rhcos.allow_missing_brew_rpms and "No such rpm" in str(e):
+                        self.logger.warning("Failed to find RPM %s in Brew", nvra, exc_info=True)
                         continue  # if conigured, just skip RPMs brew doesn't know about
                     raise Exception(f"Failed to find RPM {nvra} in brew: {e}")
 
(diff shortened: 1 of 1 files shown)


doozerlib/rhcos.py
```diff
@@ -354,7 +354,13 @@ def get_package_build_objects(self) -> Dict[str, Dict]:
         aggregate: Dict[str, Dict] = dict()
         with self.runtime.pooled_koji_client_session() as koji_api:
             for nvra in self.get_rpm_nvras():
-                rpm_def = koji_api.getRPM(nvra, strict=True)
+                try:
+                    rpm_def = koji_api.getRPM(nvra, strict=True)
+                except koji.GenericError as e:
+                    if self.runtime.group_config.rhcos.allow_missing_brew_rpms:
```
vfreex, line 360: Let's be more specific:
```suggestion
                    if self.runtime.group_config.rhcos.allow_missing_brew_rpms and "No such rpm" in str(e):
```

doozerlib/rhcos.py
```diff
@@ -354,7 +354,13 @@ def get_package_build_objects(self) -> Dict[str, Dict]:
...
+                        continue  # if conigured, just skip RPMs brew doesn't know about
```
vfreex, line 361: And get the exception logged:
```suggestion
                        self.logger.warning("Failed to find RPM %s in Brew", nvra, exc_info=True)
                        continue  # if conigured, just skip RPMs brew doesn't know about
```

vfreex (changes requested):

PR #739, opened 2023-03-08
Fix version detection error on pip installations

vfreex:
Fix the following error when invoking `doozer --version` on an installation without the source git repository:

```
LookupError: setuptools-scm was unable to detect version for /usr/local/lib/python3.11/site-packages.
```

Steps to reproduce:

```sh
pip install .

doozer --version
```

This can be solved by introducting setuptools_scm to write a file containing the to-be-installed version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ cover
 # Ignore ocp-build-dat
 ocp-build-data/
 tests_functional.log
+
+# Ignore _version.py created by setuptools_scm
+_version.py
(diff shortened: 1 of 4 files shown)


locriandev:
/lgtm

curious though to know where this was failing

vfreex:
@locriandev similar to https://github.com/openshift-eng/elliott/commit/e158e828f08470c69963efd0ec85daae4f0e72ca.

PR #743, opened 2023-03-13
Prevent "Ready" from being interpreted as a stand-in for "Accepted"

jupierce:
TRT needs us to preserve at least two `Accepted` releases payloads in the stream. For multi-streams, the release controller allows multiple `Ready` payloads at the same time in the stream. These `Ready` payloads were being interpreted as non-rejected and incorrectly satisfying the `Accepted` requirement.


joepvd (approved):
/lgtm

thiagoalessio (approved):
/lgtm

PR #750, opened 2023-03-30
gen-payload: Support weekly kernel tagging convention

vfreex:
This feature will only be enabled if `rpm_deliveries` key is defined in group config.

e.g.

```
rpm_deliveries:
  - packages:  # Enable support for "Weekly" kernel delivery through OCP
      - kernel
      - kernel-rt
    integration_tag: early-kernel-integration-8.6
    stop_ship_tag: early-kernel-stop-ship
    ship_ok_tag: early-kernel-ship-ok
```

With this config, gen-payload will do additional checks on packages (kernel and kernel-rt) to ensure they follow the following policy:

1. If any package is tagged into `early-kernel-stop-ship`, build-sync will generate an `IMPERMISSIBLE` assembly issue.
2. When running for a `standard` assembly (GA release), build-sync will generate an `IMPERMISSIBLE` assembly issue if any package doesn't have `early-kernel-ship-ok` tag.

`RHCOSBuildInspector.get_package_build_objects` is optimized to use multicall to increase the performance.


doozerlib/assembly_inspector.py
```diff
@@ -45,6 +45,19 @@ def __init__(self, runtime: Runtime, brew_session: ClientSession = None, lookup_
...
+        # Preprocess rpm_deliveries group config
+        # This is mainly to support weekly kernel delivery
+        self._rpm_deliveries: Dict[str, Dict[str, Any]] = {}  # Dict[package_name] => per package rpm_delivery config
+        if self.runtime.group_config.rpm_deliveries:
+            for entry in self.runtime.group_config.rpm_deliveries:
+                packages = entry.packages
+                if not packages:
+                    raise ValueError("`packages` in `group_config.rpm_deliveries` can't be empty")
+                if not (entry.stop_ship_tag or entry.ship_ok_tag):
+                    continue  # no validation needed
```
sosiouxme, line 57: what would be the point in having an entry like this? seems likely to be a mistake (e.g. typo)

doozerlib/assembly_inspector.py
```diff
@@ -63,12 +76,52 @@ def does_permit(self, issue: AssemblyIssue) -> bool:
...
+        :param component_description: Component description, used to include more details for the component in the AssemblyIssue message
+        :param rpm_packages: A dict; key is rpm package name, value is brew build dict
+        """
+        self.runtime.logger.info("Checking installed rpms in %s", component_description)
+        issues: List[AssemblyIssue] = []
+        for package_name, rpm_build in rpm_packages.items():
+            if package_name in self._rpm_deliveries:
+                rpm_delivery_config = self._rpm_deliveries[package_name]
+                self.runtime.logger.info("Getting tags for rpm build %s...", rpm_build['nvr'])
+                tag_names = {tag["name"] for tag in self.brew_session.listTags(build=rpm_build["id"])}
```
sosiouxme, line 95: feels like it might be a bit slow to look these tags up for every entry that has a listed RPM. could be a cached function?

doozerlib/assembly_inspector.py
```diff
@@ -63,12 +76,52 @@ def does_permit(self, issue: AssemblyIssue) -> bool:
...
+                # If the rpm is tagged into the stop-ship tag, it is never permissible
+                if rpm_delivery_config.stop_ship_tag and rpm_delivery_config.stop_ship_tag in tag_names:
+                    issues.append(AssemblyIssue(f'{component_description} has {rpm_build["nvr"]}, which has been tagged into the stop-ship tag: {rpm_delivery_config.stop_ship_tag}', component=component, code=AssemblyIssueCode.IMPERMISSIBLE))
+                    continue
+                # For GA releases, the rpm must have explicit "ship-ok" tag
```
sosiouxme, line 100: might want a permissible issue code for pre-GA releases (ECs, RCs), so that we are at least conscious we're going to need a change for GA?

sosiouxme (approved):
i'd have hoped to see test coverage increase with these changes, but i know that can be tedious.

a few other minor suggestions, but on the whole i'm good with it - merge at will

doozerlib/assembly_inspector.py
vfreex: Good point! This is a leftover from the past when kernel team hasn't started to apply stop_ship_tag and ship_ok_tag. This need to be change to raise an Error.

doozerlib/assembly_inspector.py
vfreex: Yes, I will cache this in the next push.

doozerlib/assembly_inspector.py
vfreex: We didn't really discuss what to do with pre-GA releases before. Adding a permissible issue sounds good to me.

vfreex:
Let me check how much work to add unit tests for this.

vfreex:
@sosiouxme I added some unit tests.

sosiouxme (approved):
/lgtm
great!

PR #754, opened 2023-04-03
Fix snyk errors

ashwindasr:
Changes in this PR

- Changing the name of constant from `DB_PWD` to `DB_PWD_NAME` so that synk doesn't think that its a hard coded password.
- Using `defusedxml` to fix `Insecure Xml Parser` error. Actions [run](https://github.com/openshift-eng/doozer/actions/runs/4578434428/jobs/8085035279)


sosiouxme (approved):
/lgtm

PR #758, opened 2023-04-06
fix reponame mismatch

Ximinhan:
fork repo name [openshift-bot](https://github.com/openshift-bot) different from [debug-network](https://github.com/openshift-bot/debug-network)
ref: https://redhat-internal.slack.com/archives/GDBRP5YJH/p1680734836134229?thread_ts=1680729312.015769&cid=GDBRP5YJH


ashwindasr:
The fork exists, and creating a fork will fail because it exists right.

PR #761, opened 2023-04-18
[ART-6149] Comment on pr after build

ashwindasr:
Comment on a PR after build is completed and if its not a `dry_run`. This feature is disabled by default and can be enabled by using the `--comment-on-pr` flag. While using this feature in production, add this flag only if `ASSEMBLY` is `stream`

Workflow:
- After a build is completed, we get the PR from the merge commit and check if the same comment has already been posted.
- If it hasn't it will post the comment.
- If the code fails, logger will post an error, but will not hinder other tasks, since enclosed in try-catch block

Unit tests added.

### DIFF
--- a/doozerlib/comment_on_pr.py
+++ b/doozerlib/comment_on_pr.py
@@ -1,0 +1,79 @@
+from ghapi.all import GhApi
+from doozerlib.pushd import Dir
+from dockerfile_parse import DockerfileParser
+
+
+class CommentOnPr:
+    def __init__(self, distgit_dir: str, token: str):
+        self.distgit_dir = distgit_dir
+        self.token = token
+        self.owner = None
+        self.repo = None
+        self.commit = None
+        self.gh_client = None  # GhApi client
+        self.pr_url = None
+        self.pr_no = None
+
+    def list_comments(self):
+        """
+        List the comments in a PR
+        """
+        # https://docs.github.com/rest/reference/issues#list-issue-comments
+        return self.gh_client.issues.list_comments(issue_number=self.pr_no, per_page=100)
+
+    def check_if_comment_exist(self, comment):
+        """
+        Check if the same comment already exists in the PR
+        """
+        issue_comments = self.list_comments()
+        for issue_comment in issue_comments:
+            if issue_comment["body"] == comment:
+                return True
+        return False
+
+    def post_comment(self, comment):
+        """
+        Post the comment in the PR if the comment doesn't exist already
+        """
+        # https://docs.github.com/rest/reference/issues#create-an-issue-comment
+        if not self.check_if_comment_exist(comment):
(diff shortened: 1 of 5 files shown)


doozerlib/distgit.py
```diff
@@ -1046,6 +1048,25 @@ def wait(n):
...
+                                    comment_on_pr_obj.set_github_client()
+                                    # Message to be posted to the comment
+                                    message = Template("**[ART PR BUILD NOTIFIER]** _(beta)_\n\n"
+                                                       "This PR has been included in build "
+                                                       "[$nvr](https://brewweb.engineering.redhat.com/brew/buildinfo"
+                                                       "?buildID=$build_id)"
+                                                       "for distgit *$distgit_name*. \n All builds following this will "
+                                                       "include this PR.")
+                                    comment_on_pr_obj.post_comment(message.substitute(nvr=build_info["nvr"],
+                                                                                      build_id=task_id,
```
vfreex, line 1065: `task_id` is not `build_id`. You can get build_id from `build_info['id']`.

doozerlib/comment_on_pr.py
```diff
@@ -0,0 +1,79 @@
...
+            if issue_comment["body"] == comment:
+                return True
+        return False
+
+    def post_comment(self, comment):
+        """
+        Post the comment in the PR if the comment doesn't exist already
+        """
+        # https://docs.github.com/rest/reference/issues#create-an-issue-comment
+        if not self.check_if_comment_exist(comment):
```
vfreex, line 39: `self.check_if_comment_exist` calls `self.list_comments` internally, which requires `self.pr_no` is set. However `self.pr_no` is not set until `self.get_pr_from_commit` is called.

vfreex (changes requested):

doozerlib/comment_on_pr.py
ashwindasr: Can't believe I missed that! Apologies! Thank you for pointing it out!

vfreex (approved):
Thanks. This looks good to me.

@ashwindasr Did you have a chance to test it? Or we can test it in production if you find a good time slot.

ashwindasr:
Besides the unittests that I wrote, I couldn't figure out how to test it locally since there are a lot of moving parts.

ashwindasr:
Test [success](https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/lmeyer/job/lmeyer-dev/job/dev-build%252Fcustom/91/console). Comment created in [PR](https://github.com/openshift/bond-cni/pull/52)

ashwindasr:
Needs https://github.com/openshift-eng/aos-cd-jobs/pull/3650

vfreex:
LGTM

PR #764, opened 2023-04-28
[ART-4558] Use repoquery to detect non-latest rpms

vfreex:
Currently we detect outdated rpms in images and RHCOS by comparing the NVRs of the installed rpms to the NVRs of the rpms in our `rhaos-x.y-rhel-z-candidate` Brew tag.

The current approach has 2 issues. See ART-4558 and ART-3965.

With this new approach, `repoquery` command will be used to query available RPMs in the configured build repos instead of the candidate Brew tag. This will provide more accurate detection and also be able to detect rpms that are not in rhaos repos (e.g. RHEL repos).

Note this PR only covers the `gen-payload` change. A change for `scan-sources` will be another PR.

### DIFF
--- a/doozerlib/cli/release_gen_payload.py
+++ b/doozerlib/cli/release_gen_payload.py
@@ -6,7 +6,7 @@ import sys
 import os
 import json
 from pathlib import Path
-from typing import List, Optional, Tuple, Dict, NamedTuple, Iterable, Set, Any, Callable
+from typing import List, Optional, Tuple, Dict, NamedTuple, Iterable, Set, Any, Callable, cast
 from unittest.mock import MagicMock
 
 import aiofiles
@@ -361,7 +361,7 @@ class GenPayloadCli:
         for rpm_meta in rt.rpm_metas():
             self.assembly_issues.extend(assembly_inspector.check_group_rpm_package_consistency(rpm_meta))
         if rt.assembly_type is AssemblyTypes.STREAM:
-            self.detect_non_latest_rpms(assembly_inspector)
+            await self.detect_non_latest_rpms(assembly_inspector)
         # check that images for this assembly/group are consistent with the assembly definition.
         self.detect_inconsistent_images(assembly_inspector)
 
@@ -369,7 +369,7 @@ class GenPayloadCli:
         self.detect_installed_rpms_issues(assembly_inspector)
 
         # update issues found for payload images and check RPM consistency
-        self.detect_extend_payload_entry_issues(assembly_inspector)
+        await self.detect_extend_payload_entry_issues(assembly_inspector)
 
         # If the assembly claims to have reference nightlies, assert that our payload matches them exactly.
         self.assembly_issues.extend(await PayloadGenerator.check_nightlies_consistency(assembly_inspector))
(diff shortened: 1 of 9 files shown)


vfreex:
Mostly tested locally, but didn't test it in production.

Haven't got time to add unit tests.

Needs the following PRs before merging:
- 4.14 https://github.com/openshift-eng/ocp-build-data/pull/2876
- 4.13 https://github.com/openshift-eng/ocp-build-data/pull/2877
- 4.12 https://github.com/openshift-eng/ocp-build-data/pull/2878
- 4.11 https://github.com/openshift-eng/ocp-build-data/pull/2879
- 4.10 https://github.com/openshift-eng/ocp-build-data/pull/2880
- 4.9 https://github.com/openshift-eng/ocp-build-data/pull/2881
- 4.7 https://github.com/openshift-eng/ocp-build-data/pull/2877

vfreex:
Added unit tests.

doozerlib/cli/release_gen_payload.py
```diff
@@ -587,20 +588,21 @@ def detect_extend_payload_entry_issues(self, assembly_inspector: AssemblyInspect
...
         """
 
         # record issues found in assembly list and per payload entry
-        self.assembly_issues.extend(assembly_inspector.check_rhcos_issues(payload_entry.rhcos_build))
+        rhcos_build = cast(RHCOSBuildInspector, payload_entry.rhcos_build)
+        self.assembly_issues.extend(assembly_inspector.check_rhcos_issues(rhcos_build))
         payload_entry.issues.extend(ai for ai in self.assembly_issues if ai.component == "rhcos")
 
         if self.runtime.assembly_type is AssemblyTypes.STREAM:
             # For stream alone, we want to enforce that the very latest RPMs are installed.
```
joepvd, line 604: Not your change, but this comment is misleading. We're not enforcing this, we're emitting a warning. In before times, doozer would allow `OUTDATED_RPMS_IN_STREAM_BUILD` in code. Now we have a `permits` in stream to say we want this to not be a blocker.

doozerlib/repos.py
```diff
@@ -176,6 +184,48 @@ def conf_section(self, repotype, arch=ARCH_X86_64, enabled=None, section_name=No
...
+            raise ValueError('{} is not a valid arch!')
+
+        # Attempt to get the cached result before actually querying the repo
+        cache_key = (self.name, arch)
+        with Repo._list_rpms_cache_lock:
+            cached_result = Repo._list_rpms_cache.get(cache_key)
+            if cached_result is not None:  # cache hit
+                return cached_result
+
+        # Cache miss. Will actually query the repo.
```
joepvd, line 200: Think the cache miss also should be inside the lock. Otherwise, there could be multiple queries running at the same time.

joepvd (approved):
/lgtm

Left two comments that might need addressing.

Think it is a welcome change to get informed about non-current entries in rhcos builds that are from other repositories than our own. Even if this is an area that will see some movement 🔜 

`OUTDATED_RPMS_IN_STREAM_BUILD` will only create an complaint in the log, and an annotation on the imageStream. The extended complaints will not interfere with release work.

doozerlib/repos.py
vfreex: Putting repoquery inside of the lock will prevent from multiple queries over different repos. There is a possibility to run multiple queries over the same repo, I think it is acceptable for now. Though there could still be a room to optimize this by using, such as, more fine grained locks, but I'd like to see how much time this will take in production.

doozerlib/cli/release_gen_payload.py
vfreex: `enforce` -> `verify`

PR #765, opened 2023-04-28
Comment on pr  - restrict to stream

thegreyd:


tests/test_distgit/support.py
```diff
@@ -95,8 +95,10 @@ def setUp(self):
         Define and provide mock logging for test/response
         """
         self.stream = io.StringIO()
-        logging.basicConfig(level=logging.DEBUG, stream=self.stream)
```
thegreyd, line 98: 2 tests were failing because something was breaking here

locriandev (approved):

PR #772, opened 2023-05-25
ART-6099: Add a command to support weekly kernel ingestion

vfreex:
A new command `config:tag-rpms` is added to scan RPMs (usually kernel and kernel-rt) in the integration Brew tag defined in group config then tag acceptable builds into the target Brew tag. "acceptable" here means the build is not tagged into stop_ship_tag or historically tagged into the target tag.

e.g. With the following config, Doozer will try to find latest acceptable builds of kernel and kernel-rt from Brew tag early-kernel-integration-8.6, then tag them into Brew tag rhaos-4.11-rhel-8-candidate. Additionally, all builds in tag early-kernel-stop-ship will be untagged from rhaos-4.11-rhel-8-candidate.

    rpm_deliveries:
        - packages:
            - kernel
            - kernel-rt
        integration_tag: early-kernel-integration-8.6
        stop_ship_tag: early-kernel-stop-ship
        ship_ok_tag: early-kernel-ship-ok
        target_tag: rhaos-4.11-rhel-8-candidate

This command does nothing if `rpm_deliveries` is not set in group config or an entry in `rpm_deliveries` doesn't have `target_tag`.

A dry run mode is also provided to show which builds would be tagged and untagged.

Test command:

```
doozer --debug --data-path=https://github.com/vfreex/ocp-build-data.git --group=openshift-4.12 --assembly=stream config:tag-rpms --dry-run 
```

### DIFF
--- a/doozerlib/assembly_inspector.py
+++ b/doozerlib/assembly_inspector.py
@@ -2,6 +2,7 @@ from typing import Any, List, Dict, Optional
 
 from koji import ClientSession
 from doozerlib.model import Model
+from doozerlib.rpm_delivery import RPMDeliveries, RPMDelivery
 from doozerlib.rpm_utils import parse_nvr
 
 from doozerlib import brew, util, Runtime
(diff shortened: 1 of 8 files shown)


joepvd (approved):
/lgtm

PR #773, opened 2023-05-26
Fix assembly streams parsing

vfreex:
Having `streams` overrides in assembly definition doesn't work because of wrong field is parsed in `assembly_streams_config` function.


joepvd (approved):
/lgtm

PR #775, opened 2023-06-08
Fix repoquery issue caused by yum cache

vfreex:
`repoquery` could give a wrong result from a previous invocation due to reused yum cache directory.

Similar to https://github.com/openshift-eng/aos-cd-jobs/pull/3400/, this can be fixed by running `repoquery` with a dedicated cache directory.

This PR also adds `--archlist` argument to list rpms for multi-arch.

Those 2 issues only occur on RHEL 7. repoquery on RHEL 8+ is DNF based, which works correctly without this PR.


joepvd (approved):
/lgtm

PR #778, opened 2023-06-09
temp workaround for rhsm-pulp

sosiouxme:


jupierce (approved):

sosiouxme:
did a local test with doozer that seemed to work fine.

PR #784, opened 2023-06-20
Add a switch for a fast runtime initialization (only group, logging)

locriandev:
For fast operations like `doozer config:read-releases` or `doozer config:read-assembly` we only need logging, group config and very little more. This is a proposal to add a switch (disabled by default) to return from `Runtime.initialize` right after these minimal pieces are initialized.


vfreex (approved):
LGTM.
Maybe we can have 2 initialize functions instead (one is for minimized initialization and another is for full initialization)

locriandev:
> LGTM. Maybe we can have 2 initialize functions instead (one is for minimized initialization and another is for full initialization)

Would be nice to have, but init function is looking confusing so I'd rather do that as a separate task. Refactoring it now would make this change unclear, prefer to merge as is and defer the rest

PR #789, opened 2023-07-03
Use HTTPS to access distgit

locriandev:
Custom test build: https://saml.buildvm.hosts.prod.psi.bos.redhat.com:8888/job/hack/job/dpaolell/job/custom/59/console


sosiouxme (approved):
/lgtm

lol, maybe that's all that's needed. i thought there would be at least tests needing adjustments but looking closer, seems not.

PR #793, opened 2023-07-06
Add doozer config:read-rpms

locriandev:
Ref. https://issues.redhat.com/browse/ART-6671

We need an easy way to read an RPM config and determine its characteristics. In this case, we need `arches` for microshift-sync

Example usage:
```
doozer --group=openshift-4.14 -r microshift rpms:read-config --yaml 
```


sosiouxme:
although i guess i'd prefer config:read-rpms to match the pattern of the others

PR #794, opened 2023-07-07
Fix parallel_exec invocation

locriandev:
With https://github.com/openshift-eng/doozer/pull/791, we kept only the `parallel_exec` invocation that passes `terminate_event` arg to the callable. Places that used the alternate version (with no event arg being forwarded) needed an arg placeholder in their lambda signature.

Tested with:
```
doozer --assembly=stream --group 'openshift-4.7' --latest-parent-version --images kube-rbac-proxy images:rebase --version v4.7 --release '202307061519.p?' --message 'Updating Dockerfile version and release v4.7-202307061519.p?' --no-push
```


joepvd (approved):
/lgtm

PR #797, opened 2023-07-07
Support module nvr format for picking rhel suffix

thegreyd:
Follow up to https://github.com/openshift-eng/doozer/pull/788

Right now we can't parse rhel version in nvrs like "container-selinux-2.167.0-1.module+el8.5.0+12397+bf23b712:2"
which is breaking sync-for-ci-ironic job. This should fix that


joepvd (approved):
/lgtm

PR #803, opened 2023-07-20
[config:read-releases] Fetch exact branch

thegreyd:
We're using ghapi's list_files method which calls [get_branch]( https://github.com/fastai/ghapi/blob/d2ac97ecf676d1a2f86fb14b0d0c1a48193de510/ghapi/core.py#L224)
which fetches the first matched branch with given prefix (!!)
So in my case when I pointed doozer to my fork and said fetch me openshift-4.7, 
it returned "openshift-4.7-cachito" branch.

This is pretty alarming behavior, so let's not use methods that depend on get_branch()

## Test
(I had to delete the openshift-4.7-cachito branch to unblock my failing job, but 
the original behavior should be reproducible)
- `doozer --data-path=https://github.com/thegreyd/ocp-build-data --group=openshift-4.14@disable_ptp config:read-releases --yaml`

### DIFF
--- a/doozerlib/cli/__main__.py
+++ b/doozerlib/cli/__main__.py
@@ -1533,22 +1533,10 @@ def get_releases(runtime) -> dict:
         raise DoozerFatalError('A GITHUB_TOKEN environment variable must be defined!')
     github_token = os.environ['GITHUB_TOKEN']
 
+    # TODO: data_path can be a local path in which case this will fail
     owner = runtime.data_path.split('/')[-2]
     api = GhApi(owner=owner, repo='ocp-build-data', token=github_token)
-    commitish = runtime.group_commitish
-
-    # TODO: this assumes commitish is a branch it doesn't work for tags/shasum. fix it
-    # api.git.get_ref(shasum) doesn't work
-    ref = api.git.get_ref(f'heads/{commitish}')
-    tree = api.git.get_tree(ref.object.sha).tree
-    releases_file, releases_yaml = 'releases.yml', None
-    for f in tree:
-        if f.path == releases_file:
-            releases_yaml = f
-    if not releases_yaml:
-        raise DoozerFatalError(f'Cannot find {releases_file} in {runtime.group_commitish}')
-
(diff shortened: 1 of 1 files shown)


ashwindasr:
Isn't it a good idea for us to add a new function like `get_branch_exact` like

> # Define a new class that subclasses GhApi
> class MyGhApi(GhApi):
>     def get_branch_exact(self, branch=None):
>         <match the exact branch>

vfreex:
I think probably we need to avoid using those convenience methods. `https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content` should be preferred to get file content for it.

thegreyd:
@vfreex You're right we shouldn't rely on convenience methods. I remember your errata_tool rant :)  , we should put that in art-docs for guide to api packages use. 

In this case if you see https://github.com/fastai/ghapi they have provided some convenience methods and others methods are generated from openapi spec. So I'd say let's use only the official methods (get_ref and get_tree are). I'll check out the contents endpoint, maybe it's better

doozerlib/cli/__main__.py
```diff
@@ -1535,8 +1535,19 @@ def get_releases(runtime) -> dict:
...
+    commitish = runtime.group_commitish
+
+    # TODO: this assumes commitish is a branch it doesn't work for tags/shasum. fix it
+    # api.git.get_ref(shasum) doesn't work
+    ref = api.git.get_ref(f'heads/{commitish}')
+    tree = api.git.get_tree(ref.object.sha).tree
+    releases_file, releases_yaml = 'releases.yml', None
+    for f in tree:
+        if f.path == releases_file:
+            releases_yaml = f
```
locriandev, line 1547: We could use a `break` here?

locriandev (approved):
I left a minor comment, LGTM otherwise. Thanks for catching this!

thegreyd:
repos.get_content works as expected with `ref=branch/commit/tag`, I've replaced it now.
But this breaks some of the current expectations we have for data-path like local paths.. so we should think of a long term solution for this in runtime.initialize

PR #808, opened 2023-07-31
Update print method

Ximinhan:
only print correct data to output


doozerlib/cli/__main__.py
```diff
@@ -1829,10 +1829,10 @@ def config_rhcos_src(runtime: Runtime, version, output, brew_root, arch):
         out_src_dir = out_base_dir_path.joinpath('src')
         if out_src_dir.exists():
             if out_src_dir.is_symlink():
-                print(f'Output directory already contains a symlink for {package_nvr}. Skipping.')
+                runtime.logger.info(f'Output directory already contains a symlink for {package_nvr}. Skipping.')
```
Ximinhan, line 1832: this is a message info which should not print to output

joepvd (approved):
/lgtm

PR #811, opened 2023-08-02
[ART-7273] add go wrapper script in doozer

Ximinhan:
add go wrapper script in doozerlib path named `golang_builder_FIPS_wrapper.sh`
extend add modification method to support using local path from doozer source instead of only support URL source
the new key under add modification is `doozer_source` witch take relative source path within doozer

### DIFF
--- a/doozerlib/source_modifications.py
+++ b/doozerlib/source_modifications.py
@@ -79,12 +79,14 @@ class AddModifier(object):
         :param path: Destination path to the dist-git repo.
         :param overwriting: True to allow to overwrite if path exists.
           Setting to false to prevent from accidently overwriting files from in-tree source.
+        :param doozer_source: Relative source path within doozer
 
         """
         self.source = kwargs["source"]
         self.path = kwargs["path"]
         self.overwriting = kwargs.get("overwriting", False)
         self.validate = kwargs.get("validate", None)
+        self.doozer_source = kwargs.get("doozer_source", None)
 
     def act(self, *args, **kwargs):
         """ Run the modification action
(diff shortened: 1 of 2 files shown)


doozerlib/source_modifications.py
```diff
@@ -79,12 +79,14 @@ def __init__(self, *args, **kwargs):
         :param path: Destination path to the dist-git repo.
         :param overwriting: True to allow to overwrite if path exists.
           Setting to false to prevent from accidently overwriting files from in-tree source.
+        :param doozersource: Destination path to the doozer source
```
joepvd, line 82: Think we split words with `_` if an attribute has multiple words.

doozerlib/source_modifications.py
```diff
@@ -93,38 +95,43 @@ def act(self, *args, **kwargs):
...
+            if self.validate:
+                if self.validate == "yaml":
+                    yml = yaml.safe_load(content.decode("utf-8"))  # will raise an Error if invalid
+                    if yml is None:
+                        raise IOError(f"Yaml file {source_address} is empty.")
+                else:
+                    raise ValueError("Unknown 'validate' value: {self.validate}")
+        else:  # use doozer local source
+            content = Path(Path(__file__).parent, self.doozersource).read_text()
+            source_address = self.doozersource
```
joepvd, line 129: This is not an address. Think the variable is also not strictly necessary.

doozerlib/source_modifications.py
joepvd: Note that this attribute is not used in https://github.com/openshift-eng/ocp-build-data/pull/3356

joepvd (changes requested):

joepvd (approved):
/lgtm

Let's try it :)

PR #813, opened 2023-08-04
set target version when create alignment bug

Ximinhan:
re:[ART-7462](https://issues.redhat.com/browse/ART-7462) 
as we have set jira number in alignment pr's title, CI will check if the bug is valid, we need to set correct target version to meet CI's check

### DIFF
--- a/doozerlib/cli/images_streams.py
+++ b/doozerlib/cli/images_streams.py
@@ -16,10 +16,10 @@ from jira import JIRA, Issue
 from tenacity import retry, stop_after_attempt, wait_fixed
 
 from dockerfile_parse import DockerfileParser
-from doozerlib.model import Missing
+from doozerlib.model import Model, Missing
 from doozerlib.pushd import Dir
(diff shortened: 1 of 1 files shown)


doozerlib/cli/images_streams.py
```diff
@@ -734,6 +734,7 @@ def search_issues(query):
             'project': {'key': project},
             'issuetype': {'name': 'Bug'},
             'versions': [{'name': release_version}],  # Affects Version/s
+            'customfield_12323140': [{'name': Model(runtime.gitdata.load_data(key='bug').data).jira_config.target_release[-1]}],  # Target Version
```
locriandev, line 737: Would like a comment stating that `customfield_12323140` corresponds to `Target Release` field in Jira.

locriandev (changes requested):

doozerlib/cli/images_streams.py
Ximinhan: updated

locriandev (approved):

PR #814, opened 2023-08-04
[ART-7462] set alignment bug's depends

Ximinhan:
re: [ART-7462](https://issues.redhat.com/browse/ART-7462)
accoring to https://github.com/openshift/cluster-update-keys/pull/47#issuecomment-1664730578
after creating an alignment bug CI will check if it has a bug to depend on if it's not non-GA version
we need to check if a similar bug in higher version exist and try to link it

### DIFF
--- a/doozerlib/cli/images_streams.py
+++ b/doozerlib/cli/images_streams.py
@@ -744,6 +744,10 @@ Jira mapping: https://github.com/openshift-eng/ocp-build-data/blob/main/product.
             issue = jira_client.create_issue(
                 fields
             )
+            # check depend issues and set depend to a higher version issue if ture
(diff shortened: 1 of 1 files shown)


Ximinhan:
I guess CI's check is to make sure it is backported from a higher version, but our pr and bug maybe not need to strictly follow the convention

joepvd (changes requested):
Maybe setting the valid-bug label is enough? 
Bug titles are changing here: https://github.com/openshift-eng/doozer/pull/818

Ximinhan:
> 

no, the label will be refreshed by the next CI validate check run

Ximinhan:
replaced by https://github.com/openshift-eng/art-tools/pull/12

PR #815, opened 2023-08-04
remove deprecated module

Ximinhan:
the imp module is deprecated and not used
find in test results of https://github.com/openshift-eng/doozer/actions/runs/5760509829/job/15616596716?pr=814


joepvd (approved):
/lgtm

Maybe pytest warnings should be errors

PR #817, opened 2023-08-07
[ART-4761] mirror builder image streams of ocp-private namespace

Ximinhan:
if the upstream_image exists in stream config, mirror it to ocp private by config upstream_image_mirror

### DIFF
--- a/doozerlib/cli/images_streams.py
+++ b/doozerlib/cli/images_streams.py
@@ -128,6 +128,16 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
                     print(f'For {upstream_entry_name}, would have run: {arm_cmd}')
                 else:
                     exectools.cmd_assert(arm_cmd, retries=3, realtime=True, timeout=1800)
+            # mirror builder image streams to priv namespace
+            if config.upstream_image_mirror is not Missing:
+                for upstream_image_mirror_dest in config.upstream_image_mirror:
+                    priv_cmd = f'oc image mirror {upstream_dest} {upstream_image_mirror_dest}'
+                    if runtime.registry_config_dir is not None:
+                        priv_cmd += f" --registry-config={get_docker_config_json(runtime.registry_config_dir)}"
+                    if dry_run:
+                        print(f'For {upstream_entry_name}, would have run: {priv_cmd}')
+                    else:
+                        exectools.cmd_assert(priv_cmd, retries=3, realtime=True)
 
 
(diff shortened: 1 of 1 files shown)


doozerlib/cli/images_streams.py
```diff
@@ -124,6 +124,13 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
                     print(f'For {upstream_entry_name}, would have run: {arm_cmd}')
                 else:
                     exectools.cmd_assert(arm_cmd, retries=3, realtime=True, timeout=1800)
+            # mirror builder image streams to priv namespace
+            if config.upstream_image_mirror is not Missing:
```
jupierce, line 128: The ticket requests this to be a list of destinations. It will be useful to be able to mirror to multiple destinations if we need to.

doozerlib/cli/images_streams.py
```diff
@@ -124,6 +124,13 @@ def images_streams_mirror(runtime, streams, only_if_missing, live_test_mode, dry
...
+                priv_cmd = f'oc image mirror {upstream_dest} {config.upstream_image_mirror}'
+                if dry_run:
+                    print(f'For {upstream_entry_name}, would have run: {priv_cmd}')
+                else:
+                    exectools.cmd_assert(priv_cmd, retries=3, realtime=True)
```
jupierce, line 133: Looks like we might be missing the optional registry config dir in the command line we are building up: https://github.com/openshift-eng/doozer/pull/817/files#diff-01fdff6d5a9f916a747cb4715872dca3b457bcca96c7e8a278627d522adb723bR110-R111 .

jupierce (changes requested):

doozerlib/cli/images_streams.py
Ximinhan: updated, treat it as a list in https://github.com/openshift-eng/ocp-build-data/pull/3347/files

doozerlib/cli/images_streams.py
Ximinhan: updated in https://github.com/openshift-eng/doozer/pull/817/commits/2bbe1a3cd59b5ae94d52d4d5b74238164211a593

jupierce:
@Ximinhan - this looks good. But it will need to be opened against openshift-eng/art-tools now since we migrated today.

Ximinhan:
ok, replaced by https://github.com/openshift-eng/art-tools/pull/11

PR #818, opened 2023-08-10
Use component name consistent for reconciliation

jupierce:
https://github.com/openshift-eng/ocp-build-data/blob/main/product.yml uses component name to map issues to owners. Use of distgit name in PRs/tickets has led to confusion when updating product.yml. Prevent this by using component name everywhere.


Ximinhan:
this pr needs to be reopened against [art-tools](https://github.com/openshift-eng/art-tools)

ashwindasr:
Migrated to art-tools https://github.com/openshift-eng/art-tools/pull/97

